PR #97, opened 2025-03-07
feat: workflow supports mapping from nested fields and to nested fields

shentongmartin:
#### What type of PR is this?

feat: workflow supports mapping from nested fields and to nested fields

1. can create field mapping from fields of nested structs or nested maps
2. can create field mapping to fields of nested structs or nested maps
3. use a new set of API to create nested field mapping
4. allow an intermediate level of the mapping 'source' to be an interface, will defer type check to request time: 

e.g.
```
// the output of from_node_key is map[string]any.
// at request time, the value for key 'field1' is another map with a key 'field2'
// when Compiling the workflow, we only know the value type of the map is 'any'.
// So we will defer type checking to request time.
AddInput("from_node_key", FromFieldPath([]string{"field1", "field2"})
```

### DIFF
--- a/compose/field_mapping.go
+++ b/compose/field_mapping.go
@@ -17,6 +17,7 @@
 package compose
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"runtime/debug"
@@ -58,6 +59,8 @@ func (m *FieldMapping) String() string {
 	sb.WriteString("; ")
 	return sb.String()
 }
+
+const pathSeparator = "."
 
 // FromField creates a FieldMapping that maps a single predecessor field to the entire successor input.
 // This is an exclusive mapping - once set, no other field mappings can be added since the successor input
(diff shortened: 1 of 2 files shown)


compose/field_mapping.go
```diff
@@ -59,10 +60,13 @@ func (m *FieldMapping) String() string {
 	return sb.String()
 }
 
+const pathSeparator = "."
```
meguminnnnnnnnn, line 63: 如果是map的话，key里有.怎么办

shentongmartin: Key 里不能有，有就挂了

meguminnnnnnnnn: 换成数组呢，FromField(from ...string)？

meguminnnnnnnnn: 好像不太好。。

shentongmartin: done

compose/field_mapping.go
```diff
@@ -163,26 +227,107 @@ func assignOne[T any](dest T, taken any, to string) (T, error) {
...
+		origin := reflect.New(typ)
+		inst := origin
+
+		for typ.Kind() == reflect.Ptr {
+			typ = typ.Elem()
+			inst = inst.Elem()
+			inst.Set(reflect.New(typ))
+		}
 
+		return origin
```
meguminnnnnnnnn, line 327: origin是typ的指针？  是不是少了个.Elem

compose/field_mapping.go
```diff
@@ -198,64 +343,103 @@ func checkAndExtractFromField(fromField string, input reflect.Value) (reflect.Va
...
+	if output.Kind() == reflect.Ptr {
+		if output.IsNil() {
+			originalOutput.Set(reflect.New(output.Type().Elem()))
+		}
+
 		output = output.Elem()
 	}
 
+	if output.Kind() == reflect.Ptr {
+		return reflect.Value{}, fmt.Errorf("field mapping to a struct field, but it's a nested pointer, type= %v", originalOutput.Type())
```
meguminnnnnnnnn, line 435: 为啥不允许多层指针

compose/field_mapping.go
shentongmartin: 最开始有 typ = typ.Elem()

compose/field_mapping.go
shentongmartin: 没场景，太麻烦

meguminnnnnnnnn (approved):

PR #101, opened 2025-03-12
feat: add AppendGlobalHandlers instead of InitCallbackHandlers

hi-pender:
InitCallbackHandlers always overrides the existing GlobalHandlers, which makes it difficult to inject separate Callback Handlers for different scenarios. Therefore, the AppendGlobalHandlers method has been added.

### DIFF
--- a/callbacks/interface.go
+++ b/callbacks/interface.go
@@ -52,8 +52,17 @@ type Handler = callbacks.Handler
 // InitCallbackHandlers sets the global callback handlers.
 // It should be called BEFORE any callback handler by user.
 // It's useful when you want to inject some basic callbacks to all nodes.
+// Deprecated: Use AppendGlobalHandlers instead. This function overwrites all existing global handlers.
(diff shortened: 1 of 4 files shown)


callbacks/interface.go
```diff
@@ -52,10 +52,19 @@ type Handler = callbacks.Handler
 // InitCallbackHandlers sets the global callback handlers.
 // It should be called BEFORE any callback handler by user.
 // It's useful when you want to inject some basic callbacks to all nodes.
+// Deprecated: Use AppendGlobalHandlers instead. This function overwrites all existing global handlers.
```
meguminnnnnnnnn, line 55: This function overwrites all existing global handlers.   为啥提这个，这个不构成要Deprecated的原因吧

hi-pender: 删掉了

meguminnnnnnnnn (approved):

meguminnnnnnnnn (approved):

PR #137, opened 2025-03-27
feat: introduce RegisterValuesMergeFunc to support custom merging

### LINKED ISSUE #132: [Proposal] Support for custom merge methods
Currently, eino only supports maps when merging the outputs of multiples nodes.

See the [doc](https://www.cloudwego.io/zh/docs/eino/core_modules/chain_and_graph_orchestration/orchestration_design_principles/#%E6%89%87%E5%85%A5%E4%B8%8E%E5%90%88%E5%B9%B6).

The limitation bothered me many times, it seems what I need is the "custom merge method" feature. See https://github.com/cloudwego/eino/pull/73#issuecomment-2667488838

~This proposal issue describes a way to approach that, and #137 is a proof of concept for it. If it's acceptable and there are no others working on it, I will finish the PR.~

#137 will implement it.

## Proposal

1. Define a new package function ~`func RegisterFanInMergeFunc[T any](fn func([]T) (T, error))`~ `func RegisterValuesMergeFunc[T any](fn func([]T) (T, error))`.

Users can register a custom merge function for a specific type `T`.
The function should take a slice of `T` and return a single `T` and an error.

2. Use the registered merge function first when merging the outputs of multiple nodes.

When merging the outputs of multiple nodes, check if a custom merge function is registered for the type of the output.
If so, use that function to merge the outputs. If not, fall back to the default map merging behavior.

The check for a custom merge function should take precedence over the check for map type.
So if users register a custom merge function for a map type like `type MyOutput map[string]any`,
the custom merge function should be used instead of the default map merging behavior.

## Deprecated proposal

The deprecated approach, which is based on adding new `Merge` methods for types.

#133 is a proof of concept for it

<details>
  <summary>Click to expand</summary>

1. Define a new interface with generics

```go
type Mergeable[T any] interface {
	Merge(others ...T) T
}
```

Users can implement it to make their custom type mergeable.

However, it's just a guidance for eino users.
It's an interface with generics, so it cannot be used as type assertion like common interfaces like `v.(Mergeable)`, since the right syntax is `v.(Mergeable[T])` but the T is unknown at that time.

2. Check if the type implements the interface by reflection

Duo to the limitation of Go's generics, we cannot use type assertion to check if a type implements the interface.

So we have to use reflection to check it, it's not difficult, and I believe the performance is acceptable.

3. Call the Merge method when merging

When merging outputs, if the type implements the interface, call the Merge method to merge the values.

This check is higher priority than checking if it's a map, so it means if it's a map and implements the interface,
we will call the Merge method to merge the values.

### More thoughts

1. This approach is not perfect: "A type implementing the Mergeable interface (in a wrong way) can still cannot be merged."

We expect the user to implement the Merge method like:

```go
func (MyType) Merge(others ...MyType) MyType {/*...*/}

var _ Mergeable[MyType] = (*MyType)(nil)
```

However, a bad case can be:

```go
func (MyType) Merge(others ...OtherType) OtherType {/*...*/}

var _ Mergeable[OtherType] = (*MyType)(nil)
```

Although I believe most developers will not do this, the code syntax is correct.
But the type is not mergeable since the type of the return value is not the same as the type of the receiver.
So we have to provide a function like `IsMergeable` if the type is really mergeable.

2. Registering merging functions could be another approach.

Something like:

```go
type MergeFunc[T any] func(values ...T) T

func RegisterMergeFunc[T any](f MergeFunc[T]) {
    // register the function
}
```

Then find the function by type when merging.

But I don't like this "register before use" way personally, I don't think it brings any benefit to users or developers of eino.

</details>

The key reasons to give it up are:
1. the person defining the model and the one using might be different individuals, see https://github.com/cloudwego/eino/issues/132#issuecomment-2751262182
2. Keep the same style of the implementation of contacting stream chunks, see https://github.com/cloudwego/eino/blob/cde10e9fb08cfb885de37f98ceafb9cfb750ebe9/compose/stream_concat.go#L45

Thanks @kuhahalong for pointing them out.

wolfogre:
#### What type of PR is this?

feat: A new feature

#### Check the PR title.

- [x] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level. [User docs repo](https://github.com/cloudwego/cloudwego.github.io)

引入 `RegisterValuesMergeFunc` 以支持自定义合并。

en:

This PR is an implementation of the proposal #132.

It introduces a new package-level function `RegisterValuesMergeFunc`, which allows users to register a custom merge function for a specific type.

The registered function will be used to merge the outputs of multiple nodes when fan-in.

The PR also does some minor refactoring, moving the functions used for merging that were previously scattered in `utils.go` out and reorganizing them.

zh(optional): 

这个 PR 是对提案 #132 的实现。

它引入了一个新的包级函数 `RegisterValuesMergeFunc`，允许用户为特定类型注册自定义的合并函数。

注册的函数将在多个节点扇入时，用于合并它们的输出。

该 PR 做了一些小的重构，将原本分散在 `utils.go` 里用于合并的函数移出并做了重新组织。

Close #132.

- https://github.com/cloudwego/cloudwego.github.io/pull/1292


compose/fanin_merge.go
```diff
@@ -0,0 +1,72 @@
+/*
```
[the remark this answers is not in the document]
wolfogre: ed97b72

compose/fanin_merge.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/cloudwego/eino/internal"
+)
+
+// RegisterFanInMergeFunc registers a function to merge outputs from multiple nodes when fan-in.
+// It's used to define how to merge for a specific type.
+// For maps that can be merged naturally, you don't need to register a function unless you want to customize the merge logic.
+// Please note that it's completely different from RegisterStreamChunkConcatFunc which is used to concat stream chunks.
```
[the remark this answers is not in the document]
wolfogre: 2a979e4

wolfogre:
The failed CICD doesn't seem related. It failed when "Send Coverage Report" was executed. Can anyone help?

wolfogre:
> The failed CICD doesn't seem related. It failed when "Send Coverage Report" was executed. Please help.

@meguminnnnnnnnn Could you please take a look at this?

meguminnnnnnnnn:
It's because the bot cannot comment on this PR, no need to worry about it. 

I have been quite busy recently, do you need this PR urgently? If not, I may come back to review it this weekend or early next week.

wolfogre:
It's fine, please take your time.

compose/values_merge.go
```diff
@@ -0,0 +1,71 @@
...
+package compose
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/cloudwego/eino/internal"
+)
+
+// RegisterValuesMergeFunc registers a function to merge outputs from multiple nodes when fan-in.
```
meguminnnnnnnnn, line 26: it‘s  「usually used in multiple outputs fan-in scenarios」, not only to use

wolfogre: So, could you please clarify it? "to merge outputs from multiple nodes when fan-in" and what?

I have checked the three usages of ⁠mergeValues, but all of them seem to be merging outputs.

meguminnnnnnnnn: Yes, it's only used in fan-in scenario. My original idea was not to restrict MergeValues to be only used in fan-in, so that if there are other places that need to use merge in the future, they can be used directly. However, considering that the added usage originally requires user awareness, it can also wait until the expansion of MergeValues usage scenarios.

PR #153, opened 2025-04-08
feat: support field mapping from map key with nil value or nil struct field

shentongmartin:
#### What type of PR is this?

1. can create field mapping from map key with a nil value
2. can create field mapping from struct field with a nil value

### DIFF
--- a/compose/field_mapping.go
+++ b/compose/field_mapping.go
@@ -252,7 +252,11 @@ func assignOne(destValue reflect.Value, taken any, to string) (reflect.Value, er
 					return destValue, err
 				}
 
-				destValue.SetMapIndex(key, toSet)
+				if !toSet.IsValid() {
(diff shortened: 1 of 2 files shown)


compose/field_mapping.go
```diff
@@ -252,7 +252,11 @@ func assignOne(destValue reflect.Value, taken any, to string) (reflect.Value, er
 					return destValue, err
 				}
 
-				destValue.SetMapIndex(key, toSet)
+				if toSet == (reflect.Value{}) {
```
meguminnnnnnnnn, line 255: value有isZero、isValid方法，不能用那个吗

shentongmartin: 改成 IsValid 了

meguminnnnnnnnn (approved):

PR #164, opened 2025-04-15
fix: map key not found for source field no longer results in error

shentongmartin:
#### What type of PR is this?

1. when input is stream, multiple chunks can each have in complete set of the field mappings from the pointview of successor node. If this happens, a 'MapKeyNotFound' error will happen, which is not ideal since after all chunks are processed, the field mappings may be complete.
2. So this PR removes this check: it will allow chunks to be processed even if it does not contain all field mappings.
3. To keep consistency between streaming and non-streaming scenarios, this PR removes the 'MapKeyNotFound' check out right.


compose/workflow_test.go
```diff
@@ -1067,3 +1067,22 @@ func TestNilValue(t *testing.T) {
 		assert.Equal(t, inOut{A: (*string)(nil)}, result)
 	})
 }
+
+func TestStreamFieldMap(t *testing.T) {
+	t.Run("multiple incomplete chunks in source stream", func(t *testing.T) {
```
meguminnnnnnnnn, line 1072: 加个 invoke 报错的单测吧，隐约感觉这个 diff 以后可能会酿成大祸。。。。

meguminnnnnnnnn (approved):

PR #181, opened 2025-04-23
feat: Add Unwrap method to compose internalError

cloud2303:
#### What type of PR is this?
This PR adds the `Unwrap() error` method to the `internalError` struct in the `compose` package to support Go's error unwrapping mechanism (`errors.Unwrap` and `errors.Is`). It also adds a test case to verify proper handling of `context.Canceled` errors.


meguminnnnnnnnn (approved):
thanks for your contribution~

PR #218, opened 2025-05-14
fix: support using static values for all input

shentongmartin:
#### What type of PR is this?

fix a bug when a workflow node's all inputs are static values, the nested field paths are wrong. 
Now all the field paths for the node's inputs, including nested field paths, are properly structured.

### DIFF
--- a/compose/field_mapping.go
+++ b/compose/field_mapping.go
@@ -686,19 +686,14 @@ func validateFieldMapping(predecessorType reflect.Type, successorType reflect.Ty
 		return nil, fmt.Errorf("static check fail: predecessor output type should be struct or map, actual: %v", predecessorType)
 	}
 
-	var (
-		predecessorFieldType, successorFieldType                         reflect.Type
-		err                                                              error
-		predecessorIntermediateInterface, successorIntermediateInterface bool
-	)
-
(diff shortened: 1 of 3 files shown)


compose/field_mapping.go
```diff
@@ -686,19 +686,14 @@ func validateFieldMapping(predecessorType reflect.Type, successorType reflect.Ty
...
 
-	var (
-		predecessorFieldType, successorFieldType                         reflect.Type
-		err                                                              error
-		predecessorIntermediateInterface, successorIntermediateInterface bool
-	)
-
-	for _, mapping := range mappings {
-		predecessorFieldType, predecessorIntermediateInterface, err = checkAndExtractFieldType(splitFieldPath(mapping.from), predecessorType)
+	for i := range mappings {
```
meguminnnnnnnnn, line 689: 这个为啥要这么改

meguminnnnnnnnn (approved):

compose/field_mapping.go
shentongmartin: 如果不是每次循环中重新声明变量，下面闭包函数捕获的是最后一次循环的值

PR #283, opened 2025-06-12
feat: directly print primitive type for tool output

hi-pender:
#### What type of PR is this?

directly print primitive type for tool output
don't need to use json.Marshal for primitive type

en:
zh(optional):


components/tool/utils/invokable_func.go
```diff
@@ -186,6 +186,17 @@ func (i *invokableTool[T, D]) getToolName() string {
 	return i.info.Name
 }
 
+func defaultMarshalOutput(output any) (string, error) {
+	switch v := output.(type) {
+	case string:
+		return v, nil
+	case int64, uint64, int32, uint32, int16, uint16, int8, uint8, int, uint, float64, float32, bool:
```
mrh997, line 193: int 和 float 需要单独枚举 print 吗，marshal 是不是一样的

meguminnnnnnnnn: 应该是一样的

PR #298, opened 2025-06-23
feat: can use custom extractor function in workflow field mappings

shentongmartin:
#### What type of PR is this?

Previously, workflow field mappings can only specify 'from field' or 'from path', which limits the from field or path to be along struct or map fields. 

There are cases where we want to extract from more complex paths such as array.

This PR provides a new option to specify a 'custom extractor function' when declaring a field mapping.

```
// WithCustomExtractor sets a custom extractor function for the FieldMapping.
// The extractor function is used to extract a value from the 'source' of the FieldMapping.
// NOTE: if specified in this way, Eino can only check the validity of the field mapping at request time..
func WithCustomExtractor(extractor func(input any) (any, error)) FieldMappingOption {
	return func(m *FieldMapping) {
		m.customExtractor = extractor
	}
}
```

### DIFF
--- a/compose/field_mapping.go
+++ b/compose/field_mapping.go
@@ -32,6 +32,8 @@ type FieldMapping struct {
 	fromNodeKey string
 	from        string
 	to          string
+
+	customExtractor func(input any) (any, error)
 }
 
 // String returns the string representation of the FieldMapping.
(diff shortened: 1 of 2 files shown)


compose/field_mapping.go
```diff
@@ -32,6 +32,8 @@ type FieldMapping struct {
 	fromNodeKey string
 	from        string
 	to          string
+
+	customExtractor func(input any) (any, error)
```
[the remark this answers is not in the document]
shentongmartin: FieldMapping type used to be comparable, but now it's not, because we have added a function field. To mitigate any potential break changes, we provided an Equals method for FieldMapping.

compose/field_mapping.go
```diff
@@ -548,13 +602,23 @@ func takeOne(inputValue reflect.Value, inputType reflect.Type, from string) (tak
...
+		if len(mapping.from) == 0 && mapping.customExtractor == nil {
 			return true
 		}
 	}
 	return false
 }
 
+func fromFields(mappings []*FieldMapping) bool {
+	for _, mapping := range mappings {
+		if len(mapping.from) == 0 || mapping.customExtractor != nil {
```
meguminnnnnnnnn, line 614: 这个判断是>0 && !- nil？

shentongmartin: 这个判断的意思是，这个 fieldmapping 是从具体的 field 出来，不是从 all 或者 function 出来

meguminnnnnnnnn (approved):

PR #320, opened 2025-07-07
fix: deep copy callback handlers when AppendHandlers

shentongmartin:
deep copy callback handlers when AppendHandlers, to avoid lost updates when two concurrent child processes appending to the same Context.

### DIFF
--- a/compose/utils.go
+++ b/compose/utils.go
@@ -167,6 +167,10 @@ func initGraphCallbacks(ctx context.Context, info *nodeInfo, meta *executorMeta,
 		}
 	}
 
+	if len(cbs) == 0 {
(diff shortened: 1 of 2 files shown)


compose/utils.go
```diff
@@ -167,6 +167,10 @@ func initGraphCallbacks(ctx context.Context, info *nodeInfo, meta *executorMeta,
 		}
 	}
 
+	if len(cbs) == 0 {
```
meguminnnnnnnnn, line 170: 那 init node callback 不加吗

shentongmartin: 加了，另外 len(cbs) == 0 改成调 ReuseHandlers

PR #339, opened 2025-07-15
St/adk1

shentongmartin:
#### What type of PR is this?
- make sure end user can receive from AgentEvent's MessageStream directly
- minimize MessageStream concatenation by caching the concatenated result
- SetFinalizer when creating new streams through Pipe and Copy


adk/interface.go
```diff
@@ -104,6 +148,67 @@ type AgentEvent struct {
...
+// If the MessageVariant is streaming, the MessageStream will be copied.
+// RunPath will be deep copied.
+// The result of Copy will be a new AgentEvent that is:
+// - safe to set fields of AgentEvent
+// - safe to extend RunPath
+// - safe to receive from MessageStream
+// NOTE: even if the AgentEvent is copied, it's still not recommended to modify
+// the Message itself or Chunks of the MessageStream, as they are not copied.
+// NOTE: if you have CustomizedOutput or CustomizedAction, they are NOT copied.
+func (ae *AgentEvent) Copy() *AgentEvent {
```
meguminnnnnnnnn: 是不是不需要导出

poyuzhe: 改成小写的工具函数了

PR #371, opened 2025-07-30
refactor(compose): Optimize cycle detection in DAG validation

SyedaAnshrahGillani:
This pull request optimizes the cycle detection algorithm in the validateDAG
  function by replacing the existing implementation with a more efficient
  depth-first search (DFS) based approach.

  Previously, the function used a map to store the in-degree of each node and
  then iteratively removed nodes with an in-degree of 0. This approach has a
  time complexity of O(V + E), where V is the number of vertices and E is the
  number of edges. The new implementation uses a DFS-based approach, which is
  generally faster in practice while having the same time complexity.

  Changes:

   * `compose/graph.go`: Replaced the existing validateDAG function with a more
     efficient implementation that uses a depth-first search (DFS) to detect
     cycles.

  This optimization improves the performance of the graph compilation process,
  especially for large and complex graphs.

### DIFF
--- a/compose/graph.go
+++ b/compose/graph.go
@@ -1021,53 +1021,59 @@ func transferTask(script [][]string, invertedEdges map[string][]string) [][]stri
 }
 
 func validateDAG(chanSubscribeTo map[string]*chanCall, controlPredecessors map[string][]string) error {
-	m := map[string]int{}
+	visited := make(map[string]bool)
+	recursionStack := make(map[string]bool)
+	var loops [][]string
+
 	for node := range chanSubscribeTo {
-		if edges, ok := controlPredecessors[node]; ok {
-			m[node] = len(edges)
-			for _, pre := range edges {
-				if pre == START {
-					m[node] -= 1
-				}
-			}
-		} else {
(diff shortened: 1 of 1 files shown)


shentongmartin:
Thank you for the contribution! After reviewing this carefully, we've decided not to merge this refactor for the following reasons:

1. **Regression in error reporting**: The current code reports *all* cycles in the graph (Kahn's algorithm identifies every node participating in a cycle, then `findLoops` collects all cycle paths). The DFS in this PR returns early on the first back-edge found, so only the first cycle is reported — a user with two independent cycles in their graph would only see one.

2. **No meaningful performance improvement**: Both Kahn's and DFS are O(V+E). `validateDAG` is called once at compile time on small graphs (typically tens of nodes), so there is no practical difference.

3. **The current two-phase design is intentional**: Kahn's handles detection (iterative, no recursion, deterministic); the existing `findLoops` DFS handles error formatting for confirmed cycles. Separating "is there a cycle?" from "what does it look like?" keeps the concerns clean.

Closing as declined. Thanks again for looking into this!

PR #438, opened 2025-09-02
fix: resolve edge handlers on successor ready, not on predecessor done

shentongmartin:
#### What type of PR is this?

fix an issue when a workflow node A 'indirectly maps' to another node B across branch, the field mapping on edge is eagerly resolved as soon as node A returns, which bypasses the branch control and may cause unexpected behavior.

### DIFF
--- a/compose/dag.go
+++ b/compose/dag.go
@@ -125,7 +125,8 @@ func (ch *dagChannel) reportSkip(keys []string) bool {
 	return allSkipped
 }
 
-func (ch *dagChannel) get(isStream bool) (any, bool, error) {
+func (ch *dagChannel) get(isStream bool, name string, edgeHandler *edgeHandlerManager) (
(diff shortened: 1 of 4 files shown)


compose/dag.go
```diff
@@ -125,7 +125,8 @@ func (ch *dagChannel) reportSkip(keys []string) bool {
 	return allSkipped
 }
 
-func (ch *dagChannel) get(isStream bool) (any, bool, error) {
+func (ch *dagChannel) get(isStream bool, name string, edgeHandler *edgeHandlerManager) (
```
meguminnnnnnnnn, line 128: 用 name 把 handler 取出来，再传进来，是不是好点

meguminnnnnnnnn: 也不好取，算了

PR #457, opened 2025-09-15
fix(build): compile error in Go 1.25

git-hulk:
It fails with the following errors when compiling with Go 1.25:

```
../../../../go/pkg/mod/github.com/bytedance/sonic@v1.13.3/internal/rt/stubs.go:33:22: undefined: GoMapIterator
../../../../go/pkg/mod/github.com/bytedance/sonic@v1.13.3/internal/rt/stubs.go:36:54: undefined: GoMapIterator
```

The root cause is that the sonic library doesn't work well with the new Go version. And it has been fixed in https://github.com/bytedance/sonic/issues/858, which was included in sonic 1.14.0.


git-hulk:
The CI failure is not related to this PR. Because the GitHub_TOKEN in the fork repository only has read permission. cc @meguminnnnnnnnn

PR #496, opened 2025-10-22
feat(react): add MessageRewriter configuration

shentongmartin:
### PR Description

**feat(react): Add MessageRewriter configuration**

This pull request introduces a `MessageRewriter` configuration to the `react` agent, enabling the modification of message history stored within the agent's state, without introducing any breaking changes.

**Problem:**
The existing `MessageModifier` only alters messages just before they are sent to the LLM, without affecting the agent's state. This limitation makes it difficult to implement state management features, such as compressing or summarizing the conversation history as it grows.

**Solution:**
A new `MessageRewriter` function has been added to `AgentConfig`. This provides a clear separation of concerns while maintaining backward compatibility:

*   **`MessageRewriter` (New)**: Applied at the beginning of each turn to modify the messages saved in the state. This is ideal for managing the conversation history (e.g., compression, summarization).
*   **`MessageModifier` (Unchanged)**: Its behavior remains the same. It is applied after the `MessageRewriter` and just before the LLM call. This is used for temporary modifications, like adding a system prompt that shouldn't be saved to the state.

This change provides greater flexibility for managing agent memory and state. A new test case has been added to verify that both modifiers function correctly when used together.

### DIFF
--- a/flow/agent/react/react.go
+++ b/flow/agent/react/react.go
@@ -59,8 +59,10 @@ type AgentConfig struct {
 	// modify the input messages before the model is called, it's useful when you want to add some system prompt or other messages.
 	MessageModifier MessageModifier
 
-	// MessageRewriter modifies message in the state.
-	// Useful for compressing message history to fit the model context window.
+	// MessageRewriter modifies message in the state, before the ChatModel is called.
+	// It takes the messages stored accumulated in state, modify them, and put the modified version back into state.
+	// Useful for compressing message history to fit the model context window,
+	// or if you want to make changes to messages that take effect across multiple model calls.
 	// NOTE: if both MessageModifier and MessageRewriter are set, MessageRewriter will be called before MessageModifier.
 	MessageRewriter MessageModifier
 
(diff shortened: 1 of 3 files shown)


flow/agent/react/react.go
```diff
@@ -59,6 +59,11 @@ type AgentConfig struct {
 	// modify the input messages before the model is called, it's useful when you want to add some system prompt or other messages.
 	MessageModifier MessageModifier
 
+	// MessageRewriter modifies message in the state.
```
mrh997: 注释一下调用时机？

shentongmartin: 下面一行（63）算不算

mrh997: 看到了，但感觉不太够，能理解会在 modifier 之前调用，但没有说明会在什么时候调用，类似 modifier 加个 before the model is called 就够了。

shentongmartin: 加了一些

mrh997 (approved):

PR #503, opened 2025-10-28
fix(adk): session now only store events with output

shentongmartin:
Currently events without output are also stored in session, such as Interrupt Event emitted by ChatModelAgent, which contains State ( inside of which are ever-growing list of messages). 

This is useless now because events without output can never get into an agent's actual input. 

This also introduces a potential problem of growing session storage cost since multiple interrupt events with large State might accumulate in session.

### DIFF
--- a/adk/flow.go
+++ b/adk/flow.go
@@ -387,12 +387,16 @@ func (a *flowAgent) run(
 
 		event.AgentName = a.Name(ctx)
 		event.RunPath = runCtx.RunPath
-		// copy the event so that the copied event's stream is exclusive for any potential consumer
-		// copy before adding to session because once added to session it's stream could be consumed by genAgentInput at any time
-		copied := copyAgentEvent(event)
-		setAutomaticClose(copied)
-		setAutomaticClose(event)
-		runCtx.Session.addEvent(copied)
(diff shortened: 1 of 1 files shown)


adk/flow.go
```diff
@@ -387,12 +387,15 @@ func (a *flowAgent) run(
 
 		event.AgentName = a.Name(ctx)
 		event.RunPath = runCtx.RunPath
-		// copy the event so that the copied event's stream is exclusive for any potential consumer
-		// copy before adding to session because once added to session it's stream could be consumed by genAgentInput at any time
-		copied := copyAgentEvent(event)
-		setAutomaticClose(copied)
-		setAutomaticClose(event)
-		runCtx.Session.addEvent(copied)
+		if event.Output != nil {
```
hi-pender: transfer 和 CustomizedAction 也需要存到 session 里面

shentongmartin: transfer 分两种，一种是常规的 message，一种是 transfer action。这里是去掉了 transfer action。customizedAction 这东西至少目前是没有进入上下文的

shentongmartin: 改成只去掉 interrupt event 了

PR #608, opened 2025-12-04
fix(serialization): resolve checkpoint deserialization panics on arrays and slices

### LINKED ISSUE #607: fix(serialization): checkpoint deserialization panics on arrays and uninitialized slices
## Description

The `internal/serialization/serialization.go` file contains two critical bugs that can cause panics during checkpoint deserialization, particularly when handling array-typed fields (e.g., `uuid.UUID` which is `[16]byte`) and uninitialized slices.

### Bug 1: Slice Initialization Panic in `createValueFromType`

The current logic uses `IsNil()` to determine whether a slice needs initialization:

```go
if (derefValue.Kind() == reflect.Slice || derefValue.Kind() == reflect.Array) && derefValue.IsNil() {
    derefValue.Set(reflect.MakeSlice(derefValue.Type(), 0, 0))
}
```

**Problem**

`IsNil()` can panic on uninitialized slices produced via `reflect.New().Elem()` because the slice header may be uninitialized.

**Solution**

Use a safer check for zero-initialized slices and handle arrays separately:
- For slices: check `Len() == 0 && Cap() == 0` and initialize via `MakeSlice`.
- For arrays: avoid nil checks entirely—arrays cannot be nil by definition.

### Bug 2: Incorrect Array vs Slice Handling in `setSliceElems`

The current implementation appends elements using:

```go
dResult.Set(reflect.Append(dResult, reflect.ValueOf(value)))
```

**Problem**

`reflect.Append()` does not work for arrays such as `[16]byte`, e.g., `uuid.UUID`, causing:

```
reflect: call of unknown method on array Value
```

**Solution**

Introduce explicit handling for arrays:
- Arrays: assign elements with `Index(i).Set()`.
- Slices: continue using `reflect.Append()`.

### Reproduction

Panics can be reproduced when deserializing checkpoint data containing:
- Nested structs with slice fields
- Types with fixed-size arrays like `uuid.UUID` (`[16]byte`)
- Custom structs wrapping array fields
- Checkpointed values constructed via `reflect.New().Elem()` that produce uninitialized slices

Juelz111:
## Summary

- Fix panic in `setSliceElems` when deserializing array types (e.g., `[16]byte` for UUID)
- Fix panic in `createValueFromType` when checking uninitialized slices with `IsNil()`
- Add test coverage for array and nested slice deserialization

### `setSliceElems`
Arrays cannot use `reflect.Append()`. Now distinguishes between:
- **Arrays**: Use `Index(i).Set()` for element assignment
- **Slices**: Continue using `reflect.Append()`

### `createValueFromType`
`IsNil()` can panic on uninitialized slices created via `reflect.New().Elem()`. Now uses:
- **Slices**: Check `Len() == 0 && Cap() == 0` before initialization
- **Arrays**: Skip initialization (arrays cannot be nil)

## How to Reproduce
 The issue occurs when deserializing a struct that contains an array field (like uuid.UUID, which is [16]byte). This is encountered when attempting to restore state from a checkpointer.

_**Code to Reproduce**_

```go
  package main

  import (
        "github.com/google/uuid"
        "github.com/cloudwego/eino/internal/serialization"
        "github.com/cloudwego/eino/schema"
  )

  type Item struct {
        ID uuid.UUID
  }

  func init() {
        schema.RegisterName[Item]("item")
  }

  func main() {
        s := &serialization.InternalSerializer{}

        original := Item{ID: uuid.New()}

        // Marshal works fine
        data, _ := s.Marshal(original)

        // Unmarshal panics with:
        // "reflect: call of reflect.Value.IsNil on array Value"
        var result Item
        s.Unmarshal(data, &result)
  }
```

  Root cause:

  `uuid.UUID` is defined as `[16]byte` (an array). During deserialization, two issues occur:

  1. `createValueFromType` calls `IsNil()` on array values, but arrays cannot be nil in Go, so this panics
  2. `setSliceElems` uses `reflect.Append()` which only works on slices, not arrays

  The fix distinguishes between arrays and slices, using `Index(i).Set()` for arrays instead of `Append()`, and skipping the `IsNil()` check for array types.

## Test plan

- [x] Added unit tests for array type deserialization (`TestArrayDeserialization`)
- [x] Added unit tests for nested slice/array structures
- [x] Ran `gofmt` and `golangci-lint`
- [x] All existing tests pass

Closes #607

### DIFF
--- a/internal/serialization/serialization.go
+++ b/internal/serialization/serialization.go
@@ -530,6 +530,28 @@ func internalSpecificTypeUnmarshal(is *internalStruct, typ reflect.Type) (any, e
 
 func setSliceElems(dResult reflect.Value, values []*internalStruct) error {
 	t := dResult.Type()
+
+	// Handle arrays differently from slices
+	// Arrays have fixed size and cannot use reflect.Append
+	if dResult.Kind() == reflect.Array {
+		for i, internalValue := range values {
+			if i >= dResult.Len() {
+				return fmt.Errorf("array index out of bounds: trying to set index %d in array of length %d", i, dResult.Len())
+			}
+			value, err := internalUnmarshal(internalValue, t.Elem())
+			if err != nil {
+				return fmt.Errorf("unmarshal array[%s] element %d fail: %v", t.Elem(), i, err)
+			}
+			if value == nil {
+				dResult.Index(i).Set(reflect.Zero(t.Elem()))
+			} else {
+				dResult.Index(i).Set(reflect.ValueOf(value))
+			}
+		}
+		return nil
+	}
+
+	// For slices, use Append as before
 	for _, internalValue := range values {
 		value, err := internalUnmarshal(internalValue, t.Elem())
 		if err != nil {
@@ -636,9 +658,14 @@ func createValueFromType(t reflect.Type) (value reflect.Value, derefValue reflec
 		derefValue.Set(reflect.MakeMap(derefValue.Type()))
 	}
 
-	if (derefValue.Kind() == reflect.Slice || derefValue.Kind() == reflect.Array) && derefValue.IsNil() {
-		derefValue.Set(reflect.MakeSlice(derefValue.Type(), 0, 0))
+	// Use Len() == 0 instead of IsNil() for slices to avoid panic
+	// IsNil() can panic on uninitialized slice values created via reflect.New().Elem()
+	if derefValue.Kind() == reflect.Slice {
+		if derefValue.Len() == 0 && derefValue.Cap() == 0 {
+			derefValue.Set(reflect.MakeSlice(derefValue.Type(), 0, 0))
+		}
 	}
+	// Arrays cannot be nil and don't need initialization
 
 	return value, derefValue
 }
--- a/schema/serialization_test.go
+++ b/schema/serialization_test.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/cloudwego/eino/internal/serialization"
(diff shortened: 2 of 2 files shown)


internal/serialization/serialization.go
```diff
@@ -636,9 +658,14 @@ func createValueFromType(t reflect.Type) (value reflect.Value, derefValue reflec
 		derefValue.Set(reflect.MakeMap(derefValue.Type()))
 	}
 
-	if (derefValue.Kind() == reflect.Slice || derefValue.Kind() == reflect.Array) && derefValue.IsNil() {
-		derefValue.Set(reflect.MakeSlice(derefValue.Type(), 0, 0))
+	// Use Len() == 0 instead of IsNil() for slices to avoid panic
+	// IsNil() can panic on uninitialized slice values created via reflect.New().Elem()
```
meguminnnnnnnnn, line 662: I don't quite understand this scenario. Could you please provide a piece of code that can reproduce it？

Juelz111: Yes thanks for the prompt reply. I have added another tests that shows a more direct scenario.

 The issue occurs when deserializing a struct that contains an array field (like uuid.UUID, which is [16]byte). This is encountered when attempting to restore state from a checkpointer.

_**Code to Reproduce**_

```go
  package main

  import (
        "github.com/google/uuid"
        "github.com/cloudwego/eino/internal/serialization"
        "github.com/cloudwego/eino/schema"
  )

  type Item struct {
        ID uuid.UUID
  }

  func init() {
        schema.RegisterName[Item]("item")
  }

  func main() {
        s := &serialization.InternalSerializer{}

        original := Item{ID: uuid.New()}

        // Marshal works fine
        data, _ := s.Marshal(original)

        // Unmarshal panics with:
        // "reflect: call of reflect.Value.IsNil on array Value"
        var result Item
        s.Unmarshal(data, &result)
  }
```

  Root cause:

  `uuid.UUID` is defined as `[16]byte` (an array). During deserialization, two issues occur:

  1. `createValueFromType` calls `IsNil()` on array values, but arrays cannot be nil in Go, so this panics
  2. `setSliceElems` uses `reflect.Append()` which only works on slices, not arrays

  The fix distinguishes between arrays and slices, using `Index(i).Set()` for arrays instead of `Append()`, and skipping the `IsNil()` check for array types.

PR #623, opened 2025-12-09
 fix: add thoughtSignature for schema.Toolcall to adapt Gemini3

yunhao-tech:
Problem

Gemini 3 uses thought signatures to maintain reasoning context across API calls. These signatures are encrypted representations of the model's internal thinking process. To ensure the model retains its reasoning capabilities, users must return these signatures unchanged in subsequent requests to the model.

Currently, the ToolCall struct in eino does not have a field to store this thought signature, making it impossible to properly support Gemini 3's reasoning context preservation feature.

Solution

Add a ThoughtSignature field to the ToolCall struct to store Gemini 3's thought signature. This field will be:

Serialized as thought_signature in JSON
Treated as an atomic field (not split across streaming chunks)
Properly merged during stream concatenation
Implementation

schema/message.go: Added ThoughtSignature field and updated concatToolCalls

### DIFF
--- a/schema/message.go
+++ b/schema/message.go
@@ -129,6 +129,9 @@ type ToolCall struct {
 
 	// Extra is used to store extra information for the tool call.
 	Extra map[string]any `json:"extra,omitempty"`
+
+	// ThoughtSignatures is used by Gemini 3 to maintain reasoning context across API calls.
+	ThoughtSignature string `json:"thought_signature,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


shentongmartin:
ThoughtSignature 目前不作为 ToolCall 的一级属性。Gemini3 的 ThoughtSignature 回传问题在 Gemini3 的 ChatModel 实现中已经解决。

PR #630, opened 2025-12-15
feat(schema): Make ToJSONSchema output deterministic

JonXSnow:
feat(schema): Make ToJSONSchema output deterministic

This commit ensures that the ToJSONSchema function produces a deterministic output for a given input.

Previously, when converting a ParamsOneOf struct containing a map[string]*ParameterInfo to a jsonschema.Schema , the iteration order of the map keys was not guaranteed. This resulted in the Properties of the generated schema having an inconsistent order across multiple calls. The same problem existed for nested objects in SubParams . This non-deterministic behavior could lead to cache misses in downstream systems that rely on a stable JSON representation of the schema.

To resolve this, the keys of the params and SubParams maps are now explicitly collected and sorted alphabetically before generating the schema. This guarantees that the order of properties in the resulting jsonschema.Schema is always consistent.

Additionally, a new unit test, TestToJSONSchemaDeterministic , has been added to verify this behavior. The test repeatedly calls ToJSONSchema on a complex, nested parameter structure and asserts that the marshaled JSON output is identical each time, confirming the deterministic nature of the function.

### DIFF
--- a/schema/tool_test.go
+++ b/schema/tool_test.go
@@ -17,10 +17,12 @@
 package schema
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/eino-contrib/jsonschema"
 	"github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestParamsOneOfToJSONSchema(t *testing.T) {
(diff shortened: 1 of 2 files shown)


schema/tool_test.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 CloudWeGo Authors
+ * Copyright 2025 CloudWeGo Authors
```
shentongmartin: 这个不用改吧

PR #634, opened 2025-12-19
feat(compose): persist rerun node inputs in checkpoint

shentongmartin:
### Summary

This PR enables automatic persistence of node inputs when a graph is interrupted using `WithGraphInterrupt`, allowing nodes to be correctly re-run with their original inputs when resuming from a checkpoint.

### Problem

When a graph is interrupted externally via `WithGraphInterrupt`, the interrupted nodes need to be re-run on resume. However, without persisting the original inputs, the re-run nodes would receive empty or incorrect inputs, leading to incorrect behavior.

### Solution

- **Auto-enable `persistRerunInput`**: When `WithGraphInterrupt` is used, the `taskManager` automatically enables input persistence for nodes that may need to be re-run
- **Stream input handling**: For streaming inputs, the stream is tee'd (copied) so the original input can be persisted while the node processes its copy
- **Context-based inheritance**: Subgraphs automatically inherit the behavior through Go's context propagation - no explicit configuration needed

### Key Changes

| File | Changes |
|------|---------|
| `compose/graph_manager.go` | Added `persistRerunInput` field to `taskManager`, persist `originalInput` before node execution |
| `compose/graph_run.go` | Set `persistRerunInput = true` when `WithGraphInterrupt` context is detected |
| `compose/checkpoint_test.go` | Added comprehensive tests for input persistence scenarios |

### Behavior

| Scenario | Input Persistence |
|----------|-------------------|
| `WithGraphInterrupt` used | ✅ Enabled automatically |
| `WithGraphInterrupt` NOT used | ❌ Disabled (no overhead) |
| Subgraph with parent using `WithGraphInterrupt` | ✅ Inherited via context |

### DIFF
--- a/compose/graph_compile_options.go
+++ b/compose/graph_compile_options.go
@@ -29,21 +29,10 @@ type graphCompileOptions struct {
 	serializer           Serializer
 	interruptBeforeNodes []string
 	interruptAfterNodes  []string
-	checkpointConfig     *CheckpointConfig
 
 	eagerDisabled bool
 
 	mergeConfigs map[string]FanInMergeConfig
-}
-
-// CheckpointConfig contains configuration options for checkpoint behavior.
-// This configuration is inherited by subgraphs unless explicitly overridden.
-type CheckpointConfig struct {
-	// PersistRerunInput enables persisting the original input for rerun nodes in checkpoint.
-	// When enabled, stream inputs are tee'd before consumption so they can be persisted
-	// and restored when resuming from an interrupt.
-	// When disabled (default), rerun nodes receive zero/empty values on resume for backward compatibility.
-	PersistRerunInput bool
 }
 
 func newGraphCompileOptions(opts ...GraphCompileOption) *graphCompileOptions {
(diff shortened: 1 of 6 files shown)


meguminnnnnnnnn (approved):

compose/graph_compile_options.go
```diff
@@ -29,12 +29,23 @@ type graphCompileOptions struct {
...
 	eagerDisabled bool
 
 	mergeConfigs map[string]FanInMergeConfig
 }
 
+// CheckpointConfig contains configuration options for checkpoint behavior.
+// This configuration is inherited by subgraphs unless explicitly overridden.
+type CheckpointConfig struct {
+	// PersistRerunInput enables persisting the original input for rerun nodes in checkpoint.
+	// When enabled, stream inputs are tee'd before consumption so they can be persisted
```
hi-pender: tee'd 是啥意思

shentongmartin: The tee command is a fundamental utility in Unix/Linux systems, named after the T-shaped pipe (it splits data flow like a T-junction). 

Its core function is:Read input from standard input (stdin), write it to one or more files, and simultaneously output it to standard output (stdout).

In simple terms: it lets you "copy" a command’s output to a file and still see it on the screen (or pass it to another command) at the same time—solving the problem where > redirects output to a file but hides it from the terminal.

PR #640, opened 2025-12-23
feat(adk): support shared parent session for AgentAsTool and Runner

hi-pender:
This change refines how run context and sessions are managed across agents, and ensures that sub‑agents invoked via tools (including the prebuilt Deep agent) share the same logical session while keeping event timelines isolated per run.

### What’s Changed
- Separate session state from event history
  
  - Introduced a dedicated runEvents struct to hold:
    - The main event log for a run.
    - Per‑lane event lists used for parallel workflows.
  - Simplified runSession to only manage session values ( SessionValues ) and its lock.
  - Updated runContext to carry both:
    - Session *runSession for shared logical state.
    - Events *runEvents for per‑run event tracking.
- Clarify session sharing semantics
  
  - WithSharedParentSession now only controls reuse of the parent runSession (session values).
  - Each Runner invocation always starts with a fresh runEvents , so event history is not shared across different runs.
  - Within a parallel workflow, forkRunCtx :
    - Shares the parent runSession (session values) across branches.
    - Creates a new runEvents lane per branch, preserving the existing fork/join behavior for events.
- Make AgentAsTool share the parent session by default
  
  - AgentAsTool now always runs its inner agent with shared runSession so that:
    - Session values written by the parent agent are visible to tools and sub‑agents.
    - Session values written by sub‑agents can be observed by the parent.
  - The tool‑level configuration flag for session sharing has been removed; sharing is now the default and only behavior for AgentAsTool, keeping the API surface simpler.
- Improve options documentation
  
  - Added concise comments for:
    - WithSessionValues – clarifying that it injects key‑value pairs into the runSession values.
    - WithSkipTransferMessages – clarifying that it skips transfer‑to‑agent messages when building history.
    - The internal withSharedParentSession helper – clarifying that it reuses the parent runSession when available.
- Add Deep agent coverage for session sharing
  
  - Extended adk/prebuilt/deep tests to cover a realistic sub‑agent scenario:
    - Constructed a Deep agent with a task tool and a custom sub‑agent.
    - Used a mocked ToolCallingChatModel (via internal/mock/components/model ) to trigger the task tool and route execution to the sub‑agent.
    - Asserted that the sub‑agent can see session values injected at the top‑level run, proving that session sharing works correctly through the Deep agent’s orchestration path.
### Behavior Guarantees
- Each Runner.Run :
  - Shares or creates a runSession depending on WithSharedParentSession .
  - Always uses a new runEvents instance for its own event timeline.
- Parallel branches:
  - Share the same runSession (session values).
  - Maintain separate event lanes that are merged in chronological order on join.
- AgentAsTool and the Deep prebuilt agent:
  - Sub‑agents invoked via tools participate in the same runSession as their parent agent.
  - Tests cover both low‑level run‑context behavior and the end‑to‑end Deep sub‑agent scenario.

### DIFF
--- a/adk/agent_tool_test.go
+++ b/adk/agent_tool_test.go
@@ -543,12 +543,12 @@ func (e *emitEventsAgent) Run(context.Context, *AgentInput, ...AgentRunOption) *
 	return it
 }
 
-// spyAgent captures runSession from ctx in a single nested run
+// spyAgent captures sessionValues from ctx in a single nested run
 type spyAgent struct {
 	a               Agent
 	mu              sync.Mutex
-	capturedEvents  *runEvents
-	capturedSession *runSession
+	capturedEvents  *runSession
+	capturedSession *sessionValues
 }
 
 func (s *spyAgent) Name(ctx context.Context) string        { return s.a.Name(ctx) }
@@ -563,13 +563,13 @@ func (s *spyAgent) Run(ctx context.Context, input *AgentInput, options ...AgentR
 	return s.a.Run(ctx, input, options...)
 }
 
-func (s *spyAgent) getCapturedEvents() *runEvents {
+func (s *spyAgent) getCapturedEvents() *runSession {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	return s.capturedEvents
 }
 
-func (s *spyAgent) getCapturedSession() *runSession {
+func (s *spyAgent) getCapturedSession() *sessionValues {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	return s.capturedSession
--- a/adk/call_option.go
+++ b/adk/call_option.go
@@ -58,7 +58,7 @@ func WithSkipTransferMessages() AgentRunOption {
 	})
 }
 
-// withSharedParentSession marks this run to reuse the parent runSession if available.
+// withSharedParentSession marks this run to reuse the parent sessionValues if available.
 func withSharedParentSession() AgentRunOption {
 	return WrapImplSpecificOptFn(func(o *options) {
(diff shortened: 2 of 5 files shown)


adk/agent_tool_test.go
```diff
@@ -631,26 +663,14 @@ func TestNestedAgentTool_RunPath(t *testing.T) {
 		}
 	}
 
-	for _, w := range outerRunCtx.Session.getEvents() {
```
shentongmartin, line 634: 这些断言改一下再保留比较好？

adk/interrupt_test.go
```diff
@@ -1198,6 +1198,45 @@ func withValue(value string) AgentRunOption {
 	})
 }
 
+func TestRunner_WithSharedParentSession_ReusesSession(t *testing.T) {
```
shentongmartin: 这个测试用例为啥在 interrupt_test 里面

shentongmartin (commented):
因为 runSession 中去掉了 Events 字段，这个提交对当前的 checkpoint 应该是个 break change

adk/interrupt_test.go
hi-pender: 放错位置了，我调整下

hi-pender: done

adk/agent_tool_test.go
hi-pender: 已经修正

hi-pender:
考虑到 runSession 结构体在 checkpoint 场景，需要被序列化存储，为了保证序列化数据的兼容性，不再拆分 runSession 结构体

PR #645, opened 2025-12-24
feat(adk): add shared session support for nested runner

hi-pender:
**Is your feature request related to a problem? Please describe.**

When nested agents are invoked (for example via AgentAsTool), child agents previously could not safely share the parent agent session values across boundaries, especially under concurrent workflows and checkpoint/resume. This made it hard to pass lightweight state through the agent graph without introducing custom plumbing.

**Describe the solution you'd like**

- Introduce thread-safe shared session values in `runSession` guarded by a dedicated `valuesMtx`.
- Allow child agents to reuse the parent session's `Values` and `valuesMtx` when `sharedParentSession` is enabled, both for fresh runs and for `Runner.Resume`/`ResumeWithParams`.
- Keep event recording (`Events`, `LaneEvents`) isolated per agent while sharing only the values map.

**Describe alternatives you've considered**

- Copying session values instead of sharing them, which breaks scenarios where parent and child need to observe each other's updates in real time.
- Using ad-hoc context keys or external stores to propagate state, which adds complexity and couples agents to infrastructure details.

**Additional context**

- Core implementation lives in `adk/runctx.go` (shared session values and lane handling), `adk/runner.go` (session wiring for `Run`/`Resume`), and `adk/agent_tool.go` (AgentAsTool always sharing parent session values).
- New tests cover AgentAsTool session sharing and Deep agent sub-agent session sharing to ensure correctness under nested and tool-based invocation.
- The change is backward compatible with existing gob-serialized `runSession` data. New fields are optional, and the code defensively handles missing mutex/value state when decoding older checkpoints.

### DIFF
--- a/adk/runctx.go
+++ b/adk/runctx.go
@@ -367,13 +367,17 @@ func ClearRunCtx(ctx context.Context) context.Context {
 }
 
 func ctxWithNewRunCtx(ctx context.Context, input *AgentInput, sharedParentSession bool) context.Context {
-	session := newRunSession()
+	var session *runSession
 	if sharedParentSession {
-		parentSession := getSession(ctx)
-		if parentSession != nil {
-			session.Values = parentSession.Values
-			session.valuesMtx = parentSession.valuesMtx
+		if parentSession := getSession(ctx); parentSession != nil {
+			session = &runSession{
+				Values:    parentSession.Values,
+				valuesMtx: parentSession.valuesMtx,
+			}
 		}
+	}
+	if session == nil {
+		session = newRunSession()
 	}
 	return setRunCtx(ctx, &runContext{Session: session, RootInput: input})
 }


adk/runner.go
```diff
@@ -142,12 +143,25 @@ func (r *Runner) resume(ctx context.Context, checkPointID string, resumeData map
...
 	o := getCommonOptions(nil, opts...)
+	if o.sharedParentSession {
+		parentSession := getSession(ctx)
+		if parentSession != nil {
+			runCtx.Session.Values = parentSession.Values
+			runCtx.Session.valuesMtx = parentSession.valuesMtx
+		}
+	}
+	if runCtx.Session.valuesMtx == nil {
+		runCtx.Session.Values = make(map[string]any)
```
shentongmartin: 如果不 share parent session，这里可能会从 checkpoint 中 load 出来自身的 session values map？一律重新 make 一个新的可能不对

internal/core/address.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/cloudwego/eino/internal/generic"
 )
 
-// AddressSegmentType defines the type of a segment in an execution address.
+// AddressSegmentType defines the type of segment in an execution address.
```
shentongmartin: 如果把 "a" 删了，得变成 "segments"...

hi-pender: 这个是 goland 提示语法不对，of 后面是不是不用使用名词复数

hi-pender: image

adk/runner.go
hi-pender: 是的，改正了， Values 的空值单独检查并赋值

internal/core/address.go
hi-pender: 大模型推荐使用 a segment，恢复了

PR #649, opened 2025-12-26
fix(adk): emit tool result AgentEvent after tool middlewares

shentongmartin:
# fix(adk): emit tool result AgentEvent after tool middlewares

## Problem

Tool results emitted in `AgentEvent` were captured **before** middleware processing. When middlewares modify tool outputs (e.g., content filtering, format transformation), users received the raw results instead of the processed ones.

This happened because tool results were emitted via `OnToolEnd` callbacks, which fire at the component boundary—before the middleware chain processes the output.

## Solution

Move tool result emission from callbacks into a **collector middleware** that sits at the outermost position of the middleware chain. This ensures it captures the final result after all other middlewares have processed it.

```go
// Prepend collector middleware so it wraps all user middlewares
config.ToolCallMiddlewares = append(
    []compose.ToolMiddleware{newToolResultCollectorMiddleware()},
    config.ToolCallMiddlewares...,
)
```

The collector receives the tool result sender via context (set in graph's `OnStartFn`), and emits the result after `next()` returns.

## Key Design Decisions

**Why context instead of state for sender passing?**

In interrupt/resume scenarios, `OnStartFn` is called before state is restored from checkpoint. Using context ensures the sender is available regardless of state restoration timing.

## Scope

Both `adk.ChatModelAgent` and `flow/agent/react.MessageFuture` are fixed with the same pattern. Tests cover both Invoke and Stream paths, verifying that middleware-modified results are correctly emitted.


adk/chatmodel.go
```diff
@@ -41,6 +41,14 @@ import (
 	ub "github.com/cloudwego/eino/utils/callbacks"
 )
 
+const (
+	addrDepthChain      = 1
+	addrDepthReactGraph = 2
+	addrDepthChatModel  = 3
```
hi-pender, line 47: 有重复编号

hi-pender: 哦哦，不是编号，是长度值

adk/react.go
```diff
@@ -99,6 +132,59 @@ func popToolGenAction(ctx context.Context, toolName string) *AgentAction {
...
+					sender = senders.sender
+					senderAddr = senders.addr
+				}
+				output, err := next(ctx, input)
+				if err != nil {
+					return nil, err
+				}
+				prePopAction := popToolGenAction(ctx, input.Name)
+				if sender != nil {
+					currentAddr := core.GetCurrentAddress(ctx)
```
hi-pender: ToolMiddleware 是直接加到指定的 ToolsNode 节点上的， 不像 Callback 生效范围会随着 graph 的变更而动态变化。 这里是不是不需要校验 isAddressAtDepth 了

shentongmartin: 实际上不需要，算是一个额外的防御

flow/agent/react/option.go
```diff
@@ -144,17 +144,34 @@ func WithMessageFuture() (agent.AgentOption, MessageFuture) {
...
 	}
 	graphHandler := callbacks.NewHandlerBuilder().
-		OnStartFn(h.onGraphStart).
-		OnStartWithStreamInputFn(h.onGraphStartWithStreamInput).
+		OnStartFn(func(ctx context.Context, info *callbacks.RunInfo, input callbacks.CallbackInput) context.Context {
+			h.onGraphStart(ctx, info, input)
+			return setToolResultSendersToCtx(ctx, createToolResultSender(), createStreamToolResultSender())
+		}).
+		OnStartWithStreamInputFn(func(ctx context.Context, info *callbacks.RunInfo, input *schema.StreamReader[callbacks.CallbackInput]) context.Context {
+			h.onGraphStartWithStreamInput(ctx, info, input)
```
hi-pender, line 168: 这里面的 input，我看没有地方用，也没有地方 close

shentongmartin: done

adk/react.go
shentongmartin: 去掉了

hi-pender (approved):

PR #672, opened 2026-01-09
fix(adk): exit properly skip deterministic transfer and enable interrupt/resume for wrapped flowAgents

shentongmartin:
## Summary

This PR fixes two issues in `AgentWithDeterministicTransferTo`:

1. **Exit events from sub-agents were being ignored** - Transfer events were unconditionally appended, overriding Exit semantics
2. **flowAgent wrapped by deterministic transfer couldn't interrupt/resume** - Using `ClearRunCtx` caused event list loss and SessionValues to be cleared

### Exit Scoping: "Exit as Last Event" Heuristic

Instead of complex runner boundary tracking, we use a simple heuristic:
- If Exit is the **last event** in the stream → honor it, skip Transfer
- If Exit is followed by other events (e.g., tool results from agentTool) → it's from a nested scope, continue with Transfer

```go
var lastEvent *AgentEvent
for {
    event, ok := iter.Next()
    if !ok {
        break
    }
    generator.Send(event)
    lastEvent = event
}

if lastEvent != nil && lastEvent.Action != nil {
    if lastEvent.Action.Interrupted != nil || lastEvent.Action.Exit {
        return  // Skip transfer
    }
}
sendTransferEvents(generator, toAgentNames)
```

### RunPath: Set Exactly Once

Simplified RunPath handling with clear ownership:
- **flowAgent**: Sets RunPath only when `len(event.RunPath) == 0`
- **agentTool**: Prepends parent RunPath when forwarding events

This works because `RunStep.agentName` is unexported - users cannot manually set RunPath.

### Isolated Session for flowAgent

For flowAgent, create an isolated session that:
- **Shares** `Values` and `valuesMtx` with parent (preserves SessionValues)
- **Isolates** `Events` (for proper interrupt/resume)

On interrupt, save the isolated session's events to `deterministicTransferState` via `CompositeInterrupt`. On resume, restore from this state.

## Backward Compatibility

- No API changes
- No breaking changes to checkpoint format

### DIFF
--- a/adk/agent_tool.go
+++ b/adk/agent_tool.go
@@ -67,6 +67,25 @@ func withAgentToolEnableStreaming(enabled bool) tool.Option {
 }
 
 // NewAgentTool creates a tool that wraps an agent for invocation.
+//
+// Event Streaming:
+// When EmitInternalEvents is enabled in ToolsConfig, the agent tool will emit AgentEvent
+// from the inner agent to the parent agent's AsyncGenerator, allowing real-time streaming
+// of the inner agent's output to the end-user via Runner.
+//
+// Note that these forwarded events are NOT recorded in the parent agent's runSession.
+// They are only emitted to the end-user and have no effect on the parent agent's state
+// or checkpoint. The only exception is Interrupted action, which is propagated via
+// CompositeInterrupt to enable proper interrupt/resume across agent boundaries.
+//
+// Action Scoping:
+// Actions emitted by the inner agent are scoped to the agent tool boundary:
+//   - Interrupted: Propagated via CompositeInterrupt to allow proper interrupt/resume across boundaries
+//   - Exit, TransferToAgent, BreakLoop: Ignored outside the agent tool; these actions only affect
+//     the inner agent's execution and do not propagate to the parent agent
+//
+// This scoping ensures that nested agents cannot unexpectedly terminate or transfer control
+// of their parent agent's execution flow.
 func NewAgentTool(_ context.Context, agent Agent, options ...AgentToolOption) tool.BaseTool {
 	opts := &AgentToolOptions{}
 	for _, opt := range options {
--- a/adk/deterministic_transfer.go
+++ b/adk/deterministic_transfer.go
@@ -137,10 +137,8 @@ func forwardEventsAndAppendTransfer(iter *AsyncIterator[*AgentEvent],
 		lastEvent = event
 	}
 
-	if lastEvent != nil && lastEvent.Action != nil {
-		if lastEvent.Action.Interrupted != nil || lastEvent.Action.Exit {
-			return
-		}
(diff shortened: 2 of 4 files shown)


adk/agent_tool.go
```diff
@@ -170,6 +189,12 @@ func (at *agentTool) InvokableRun(ctx context.Context, argumentsInJSON string, o
 
 		if gen != nil {
 			if event.Action == nil || event.Action.Interrupted == nil {
+				if parentRunCtx := getRunCtx(ctx); parentRunCtx != nil && len(parentRunCtx.RunPath) > 0 {
```
meguminnnnnnnnn, line 192: 这个为什么不在 flow agent 里做，这个 event 原本 run path 就不对吗

adk/deterministic_transfer.go
```diff
@@ -82,63 +92,187 @@ func (a *resumableAgentWithDeterministicTransferTo) Name(ctx context.Context) st
...
+		event, ok := iter.Next()
+		if !ok {
+			break
+		}
+		generator.Send(event)
+		lastEvent = event
+	}
+
+	if lastEvent != nil && lastEvent.Action != nil {
+		if lastEvent.Action.Interrupted != nil || lastEvent.Action.Exit {
```
meguminnnnnnnnn: 两个 if 为什么不写到一起

adk/deterministic_transfer.go
```diff
@@ -82,63 +92,187 @@ func (a *resumableAgentWithDeterministicTransferTo) Name(ctx context.Context) st
...
+
+	isolatedSession := &runSession{
+		Values:    parentSession.Values,
+		valuesMtx: parentSession.valuesMtx,
+	}
+	if isolatedSession.valuesMtx == nil {
+		isolatedSession.valuesMtx = &sync.Mutex{}
+	}
+	if isolatedSession.Values == nil {
+		isolatedSession.Values = make(map[string]any)
```
meguminnnnnnnnn, line 161: 不是Isolated 吗，这样不是没Isolated 吗

adk/agent_tool.go
shentongmartin: 之前是在 flow agent 里面做的 append，有个问题，是 flow->deterministic_transfer->flow，这样会 append 两次。

现在 flow 里面改成只有 RunPath 为空才会设置，append 改到 AgentTool 这里，避免上面那个情况

adk/deterministic_transfer.go
shentongmartin: 只 isolate event 列表，SessionValues 不隔离，跟 AgentTool 的 runner 行为一致

adk/deterministic_transfer.go
shentongmartin: done

meguminnnnnnnnn (approved):

PR #790, opened 2026-02-13
fix(adk/filesystem): fix emptyStreamConcatErr when streaming execute command has no output

hi-pender:
## Problem

When using `ExecuteStreaming` to run commands that produce no stdout output (e.g., `mkdir`, `touch`), the operation fails with `concat stream reader fail: stream reader is empty, concat fail` error, even though the command itself succeeds.

## Root Cause

**Call chain:**
1. `local.ExecuteStreaming` executes the command and sends `&filesystem.ExecuteResponse{ExitCode: new(int)}` to indicate successful completion (ExitCode=0) when there's no stdout
2. `newStreamingExecuteTool` calls `convExecuteResponse(chunk)` for each chunk
3. `convExecuteResponse` returns an empty string `""` because `Output=""`, `ExitCode=0` doesn't match the `!= 0` condition, and `Truncated=false`
4. Since `str == ""`, the streaming logic skips this chunk and sends nothing
5. `concatStreamReader` receives only EOF with an empty items slice, triggering `emptyStreamConcatErr`

## Solution

Refactored the streaming logic in `newStreamingExecuteTool`:

1. **During stream consumption:**
   - Track `exitCode` across chunks
   - Combine `Output` and `Truncated` info within the same chunk before sending (avoiding multiple sends per chunk)
   - Track whether any content has been sent

2. **After stream completion:**
   - If `exitCode != 0`, append failure message with error code
   - If no content was sent, send `[Command executed successfully with no output]`

3. **Preserved `convExecuteResponse`:** Used by non-streaming `newExecuteTool`, with the same fix applied

## Changes

- `adk/middlewares/filesystem/filesystem.go`: Refactored streaming logic
- `adk/middlewares/filesystem/filesystem_test.go`: Added test case for "successful command with no output"

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -37,7 +37,7 @@ jobs:
         # https://golangci-lint.run/
         uses: golangci/golangci-lint-action@v9.2.0
         with:
-          version: v2.8.9
+          version: v2.8.0
           args: --timeout 5m
 
   commit-msg-check:


adk/middlewares/filesystem/filesystem.go
```diff
@@ -407,20 +407,45 @@ func newStreamingExecuteTool(sb filesystem.StreamingShellBackend, desc *string)
...
 					sw.Send("", recvErr)
-					break
+					return
+				}
+
+				if chunk == nil {
+					continue
+				}
+				if chunk.ExitCode != nil {
+					exitCode = chunk.ExitCode
```
meguminnnnnnnnn, line 428: 这啥意思，直接取值不行吗

hi-pender: 这个是为了保留最近的非 nil 值。 chunk 流中，如果最后一个 chunk 是 nil 值，前面有非 nil 值，以最后出现的非 nil 值为准

PR #850, opened 2026-03-09
refactor(filesystem): optimize Read to return FileContent and avoid allocations

JonXSnow:
refactor(filesystem): optimize Read to return FileContent and avoid allocations

- Change Backend.Read return type from string to FileContent struct for extensibility
- Optimize InMemoryBackend.Read to use strings.IndexByte scanning instead of
  strings.Split/Join, eliminating unnecessary allocations for large files
- Add fast path for the common case (no offset, content within limit)
- Fix default limit from 200 to 2000 to match ReadRequest documentation
- Fix offset to be strictly 1-based as documented in ReadRequest
- Move line number formatting from backend to middleware layer (separation of concerns)
- Add comprehensive Read edge case tests

### DIFF
--- a/adk/filesystem/backend_inmemory.go
+++ b/adk/filesystem/backend_inmemory.go
@@ -133,7 +133,7 @@ func mustParseTime(s string) time.Time {
 }
 
 // Read reads file content with offset and limit.
-func (b *InMemoryBackend) Read(ctx context.Context, req *ReadRequest) (string, error) {
+func (b *InMemoryBackend) Read(ctx context.Context, req *ReadRequest) (*FileContent, error) {
 	b.mu.RLock()
 	defer b.mu.RUnlock()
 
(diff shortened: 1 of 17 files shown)


adk/filesystem/backend_inmemory.go
```diff
@@ -133,46 +133,61 @@ func mustParseTime(s string) time.Time {
...
 	}
 
-	lines := strings.Split(entry.content, "\n")
-	totalLines := len(lines)
+	content := entry.content
 
-	if offset >= totalLines {
-		return "", nil
+	// Fast path: no offset, content fits within limit — return as-is
+	if offset == 0 {
```
hi-pender, line 160: 这个 Read 的实现方式为啥要调整，性能是不是差不少

JonXSnow: 避免  strings.Split 全量分割与 strings.Join 重新拼接 多次操作

adk/middlewares/filesystem/filesystem.go
```diff
@@ -561,11 +561,26 @@ func newReadFileTool(fs filesystem.Backend, name string, desc string) (tool.Base
...
 			Limit:    input.Limit,
 		})
+		if err != nil {
+			return "", err
+		}
+
+		startLine := input.Offset
+		lines := strings.Split(fileCt.Content, "\n")
+		var b strings.Builder
+		for i, line := range lines {
```
hi-pender, line 580: 直接区分下最后一行，是不是更好。别再 TrimSuffix 了

PR #943, opened 2026-04-09
fix: replace direct log.Printf with injectable Logger in ModelRetryConfig

### LINKED ISSUE #938: 关于adk/retry_chatmodel.go中log输出的问题
当前`adk/retry_chatmodel.go`直接使用内置log来输出日志，这样会干扰用户终端的输出。

建议：
1. 注入 Logger 接口
```go
// Logger is the interface for logging retry events.
// If not set, retry events are silently dropped.
type Logger interface {
    Printf(format string, args ...any)
}

// ModelRetryConfig configures retry behavior for the ChatModel node.
type ModelRetryConfig struct {
    MaxRetries  int
    IsRetryAble func(ctx context.Context, err error) bool
    BackoffFunc func(ctx context.Context, attempt int) time.Duration

    // Logger is used to log retry events.
    // If nil, retry events are silently dropped.
    Logger Logger
}
```
2. 直接移除

mango766:
Fixes #938.

`retry_chatmodel.go` uses `log.Printf` directly for retry events, which outputs to the user's terminal and interferes with TUI applications.

## Changes

- Add `Logger` interface field to `ModelRetryConfig` with `Printf(format string, v ...any)` method
- When `Logger` is nil (default), retry events are silently dropped
- Replace all 3 `log.Printf` calls with `r.logf()` helper
- Remove unused `log` import

1 file changed: `adk/retry_chatmodel.go`

## Backward compatible

Default behavior changes from logging to silent. Users who want logging can set `Logger: log.Default()`.

### DIFF
--- a/adk/retry_chatmodel.go
+++ b/adk/retry_chatmodel.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"math/rand"
 	"time"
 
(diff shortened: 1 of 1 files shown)


shentongmartin:
thanks for the contribution. We probably will remove the log outright because user can always embed their own custom logging in the IsRetryAble callbacks, etc.

Issue #431, opened 2025-08-28
[Repost] [GraphRunError] failed to submit tasks: run node[tools] pre processor fail


fromstartosun:
I call react.Agent Stream() method, and ModelCallbackHandler.OnEndWithStreamOutput() output below error,
`[GraphRunError] failed to submit tasks: run node[tools] pre processor fail: concat stream reader fail: cannot concat ToolCalls with different tool id: 'tooluse_itm_mYS7TaC51v1wi7H1zQ' 'tooluse_9lc9-a5eSsGuGmKQJv9EWA'`

Before the error,
CallbackOutput.Message.Content - `Now let me check the dishes available at the top Beijing restaurants to find spicy options for you:`
CallbackOutput.Message.ToolCalls.Function.Name - `query_dishes`
CallbackOutput.Message.ToolCalls.Function.Arguments - `{"restaurant_id": "1001", "topn": 10}`
CallbackOutput.Message.ToolCalls.Function.Name - `query_dishes`
CallbackOutput.Message.ToolCalls.Function.Arguments - `{"restaurant_id": "1002", "topn": 10}`

@meguminnnnnnnnn any thought? Thanks!

shentongmartin:
check the 'index' of the two different tool calls

fromstartosun:
> check the 'index' of the two different tool calls

@shentongmartin  The first tool call index is 0, and the second one is 1.

mrh997:
@fromstartosun Did you write the concat logic yourself? If so, check that the index assignment is correct. Note that index is a pointer.

PR #962, opened 2026-04-14
fix(schema): handle different tool IDs at same index in concatToolCalls

Ed1s0nZ:
## Summary

Some OpenAI-compatible providers (e.g. **qwen**, **deepseek**) reuse the same streaming chunk `Index` for different tool calls, distinguished only by `ID`. The current `concatToolCalls` groups chunks solely by `Index`, causing:

```
[NodeRunError] cannot concat ToolCalls with different tool id: 'toolu_xxx' 'toolu_yyy'
```

This PR fixes `concatToolCalls` to detect when an `Index` group contains multiple distinct non-empty IDs and splits them into separate sub-groups, each merged independently. Chunks with empty IDs follow streaming order (assigned to the most recently seen non-empty ID). **Behavior is unchanged** for providers that correctly assign unique indexes per tool call (e.g. OpenAI).

## Root Cause

In `schema/message.go`, `concatToolCalls` uses `map[int][]int` keyed by `Index` only. When provider X sends:

```
chunk 1: Index=0, ID="tool_a", Name="nmap" 
chunk 2: Index=0, ID="tool_b", Name="sqlmap"
```

Both land in the same group, and the ID mismatch check fails.

## Fix

- Within each `Index` group, check if multiple distinct non-empty IDs exist
- If yes, split into sub-groups by ID (empty-ID chunks follow the last seen non-empty ID)
- Reassign sequential indexes to all sub-groups
- Existing single-ID-per-index behavior is preserved (no sub-grouping needed)

## Test

- Updated `different_tool_id` test case to verify two tool calls with different IDs at the same index are correctly split into separate merged results
- All existing `TestConcatToolCalls` sub-tests pass

Fixes #927 #431 #361

### DIFF
--- a/schema/message.go
+++ b/schema/message.go
@@ -1438,6 +1438,8 @@ func mergeToolTextParts(group []ToolOutputPart) (ToolOutputPart, error) {
 
 func concatToolCalls(chunks []ToolCall) ([]ToolCall, error) {
 	var merged []ToolCall
+
+	// First pass: group chunks by Index.
 	m := make(map[int][]int)
 	for i := range chunks {
 		index := chunks[i].Index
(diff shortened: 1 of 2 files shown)


river1987:
碰到同样的问题了，啥时候能合并呢

PR #966, opened 2026-04-14
feat(middlewares): add permission middleware for tool call gating

shentongmartin:
## Summary

| Problem | Solution |
|---------|----------|
| No built-in way to gate tool execution behind permission checks in the ADK middleware layer | Implement a `permission.Middleware` that intercepts tool calls with a user-defined `BeforeToolCall` evaluator, supporting Allow/Deny/Ask decisions with interrupt/resume for interactive approval |

## Key Insight: Framework Skeleton vs. Strategy Pattern

Inspired by Claude Code's permission system (7 modes, 8 rule layers, 5 parallel racers), but deliberately scoped as a **framework skeleton** rather than an opinionated strategy engine. The middleware provides the interrupt/resume plumbing; users supply the evaluation logic. This matches Eino's position as a general-purpose framework that must serve diverse product forms.

The critical design choice is the three-outcome model:
- **Allow**: Execute the tool (optionally with modified input via `UpdatedInput`)  
- **Deny**: Return a formatted deny message as the tool result—crucially, this does NOT interrupt the agent loop. The LLM sees "Permission denied" and self-adapts, matching CC's battle-tested pattern.
- **Ask**: Trigger `StatefulInterrupt` to pause the agent loop, checkpoint state, and await external approval via `ResumeWithParams`.

### Why fail-closed for unknown decisions?
Unknown `Decision` values (e.g., a typo like `"alow"`) produce a deny result rather than an error. This prevents fail-open security vulnerabilities—a permission system that accidentally allows unknown states is worse than one that blocks them.

### Why `error` only for infrastructure failures?
`BeforeToolCall` returning `error` is reserved for infrastructure failures (rule store unreachable, network timeout). This aborts the agent loop. Permission denials use `Decision: Deny` instead, which keeps the agent loop running. This separation prevents conflating policy decisions with system failures.

### Why re-interrupt on non-target resume?
When the runtime resumes a different tool but this tool was previously interrupted, it must re-interrupt to preserve its pending state. This supports parallel tool call scenarios where only one tool is approved at a time.

## What Changed

| File | Change |
|------|--------|
| `adk/middlewares/permission/permission.go` | New package: `Middleware` struct, `permissionGate` logic, `WrapInvokableToolCall`/`WrapStreamableToolCall`, all supporting types (`Decision`, `ToolCallDecision`, `AskInfo`, `AskState`, `ResumeResponse`) |
| `adk/middlewares/permission/permission_test.go` | 22 unit tests covering Allow/Deny/Ask/Unknown/Nil/Error paths for both invokable and streamable tool calls |

## How to Verify

```bash
go test -race -count=1 ./adk/middlewares/permission/...
golangci-lint run ./adk/middlewares/permission/...
```

---

## 摘要

| 问题 | 解决方案 |
|------|----------|
| ADK 中间件层缺少内置的工具调用权限控制机制 | 实现 `permission.Middleware`，通过用户定义的 `BeforeToolCall` 评估函数拦截工具调用，支持 Allow/Deny/Ask 三种决策，Ask 决策通过中断/恢复机制实现交互式审批 |

## 核心洞察：框架骨架 vs. 策略引擎

受 Claude Code 权限系统（7 种模式、8 层规则、5 路并行竞争）启发，但刻意定位为**框架骨架**而非固化的策略引擎。中间件提供中断/恢复的管道设施，用户自行实现评估逻辑。这符合 Eino 作为通用框架需服务多样产品形态的定位。

三态结果模型的关键设计：
- **Allow**：执行工具（可选通过 `UpdatedInput` 修改输入）
- **Deny**：返回格式化的拒绝消息作为工具结果——不中断 agent 循环。LLM 看到 "Permission denied" 后自行适应，沿用 CC 经过实战验证的模式。
- **Ask**：触发 `StatefulInterrupt` 暂停 agent 循环、持久化状态、等待外部通过 `ResumeWithParams` 回注审批结果。

### 为何对未知决策采用 fail-closed？
未知的 `Decision` 值产生 deny 结果而非 error，防止 fail-open 安全漏洞。

### 为何 error 仅用于基础设施故障？
`BeforeToolCall` 返回 error 仅表示基础设施故障（如规则存储不可达），会中断 agent 循环。权限拒绝使用 `Decision: Deny`，保持 agent 循环正常运行，避免将策略决策与系统故障混淆。

### 为何非目标恢复时重新中断？
当运行时恢复其他工具而当前工具之前已中断时，必须重新中断以保留挂起状态，支持并行工具调用场景下的逐个审批。


adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+	Message      string
+	UpdatedInput string
+	Reason       string
+}
+
+// BeforeToolCall is the user-provided evaluation function invoked before each tool call.
+// It returns a ToolCallDecision that determines whether the call is allowed, denied, or
+// requires interactive approval. Returning an error signals an infrastructure failure
+// and aborts the agent loop; permission denials should use Decision: Deny instead.
+type BeforeToolCall func(ctx context.Context, toolName string, argumentsInJSON string) (*ToolCallDecision, error)
```
shentongmartin: The name 'BeforeToolCall' feel a bit ambiguous as it overlaps with the handler function names defined in 'ChatModelAgentMiddleware'. Try change into more concrete and specialized name.

Also, are the 'toolName' and 'argumentsInJSON' sufficient for the function as input parameter? Do they work well with the Enhanced tools? Why not reuse ToolContext?

adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+
+// AskState is the interrupt state persisted via CheckPointStore (gob serialization).
+type AskState struct {
+	Info *AskInfo
+}
+
+// ResumeResponse is the decision injected externally via ResumeWithParams.
+type ResumeResponse struct {
+	Approved     bool
+	UpdatedInput string
```
shentongmartin, line 76: Does it work well with the EnhancedTool which seems to have structured tool input?

adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+	Message   string
+}
+
+// AskState is the interrupt state persisted via CheckPointStore (gob serialization).
+type AskState struct {
+	Info *AskInfo
+}
+
+// ResumeResponse is the decision injected externally via ResumeWithParams.
+type ResumeResponse struct {
```
shentongmartin, line 74: In fact we probably will need richer 'ResumeResponse'. For example:

- Yes, and don't ask again for git * commands in /project
- Yes, allow all edits during this session

It may be better if we provide a 'ResumeHandler' type of callback function that allows users to capture the user provided ResumeResponse and trigger their own logic.

Also, the real thing that get into the choice is ultimately decided by the 'Tool' and the 'Product' built with this permission middleware. In the 'AskInfo' struct, we MUST allow certain degree of customization by the 'Tool' itself. Such as allow a per-tool level configuration in this permission middleware.

adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Middleware is a ChatModelAgentMiddleware that gates tool execution behind
+// a user-defined BeforeToolCall permission check.
+type Middleware struct {
+	*adk.BaseChatModelAgentMiddleware
+	beforeToolCall BeforeToolCall
+}
+
+// NewMiddleware creates a new permission Middleware with the given BeforeToolCall evaluator.
+func NewMiddleware(beforeToolCall BeforeToolCall) *Middleware {
```
shentongmartin: It seems to me that OTHER ChatModelAgentMiddleware implementations are using 'New' instead of 'NewMiddleware'?

adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+		return &gateResult{allowed: true, updatedInput: input}, nil
+	}
+
+	if isTarget && !hasData {
+		return nil, fmt.Errorf("permission: tool %s was targeted for resume but received nil or type-mismatched ResumeResponse", tCtx.Name)
+	}
+
+	decision, err := m.beforeToolCall(ctx, tCtx.Name, argumentsInJSON)
+	if err != nil {
+		return nil, fmt.Errorf("permission check failed: %w", err)
```
shentongmartin: need better, more informative error msg. Remember that these error message may get into the LLM in the next agent iteration. They should serve as guidance not only to human users but to the LLMs too.

adk/middlewares/permission/permission.go
```diff
@@ -0,0 +1,208 @@
...
+	endpoint adk.StreamableToolCallEndpoint,
+	tCtx *adk.ToolContext,
+) (adk.StreamableToolCallEndpoint, error) {
+	return func(ctx context.Context, argumentsInJSON string, opts ...tool.Option) (*schema.StreamReader[string], error) {
+		result, err := m.permissionGate(ctx, tCtx, argumentsInJSON)
+		if err != nil {
+			return nil, err
+		}
+		if !result.allowed {
+			sr, sw := schema.Pipe[string](1)
```
shentongmartin: if it better to directly use StreamReaderFromArray?

adk/middlewares/permission/permission_test.go
shentongmartin, line 1: It seems to be we lack end-to-end ask -> resume test cases? Including single tool and current tools.

PR #1019, opened 2026-05-13
feat(adk): add Runner-managed session persistence with model timeout

shentongmartin:
# Runner-Managed Session Persistence

## Problem

The ADK Runner had no built-in mechanism for persisting conversation state across turns. Users who needed multi-turn persistence had to implement custom checkpoint management outside the Runner, handling serialization, interrupt/resume coordination, and session lifecycle themselves. This was error-prone—especially around race conditions during managed interrupt resume, partial-output replay, and reduction persistence.

Additionally, model calls had no timeout control. A hanging LLM connection (slow first token, stalled stream, or unbounded total duration) would block indefinitely with no way to detect or recover.

### Session Persistence

A typed `SessionService[M]` manages the full lifecycle of session state within the Runner. It uses an append-only event log (`SessionStore`) as the persistence primitive, with cursors for incremental loading and configurable page sizes with retry-backoff.

Key capabilities:
- **Session rollback**: revert to a prior checkpoint when a turn produces bad results
- **Managed interrupt/resume**: the Runner coordinates interrupt checkpoints and resume resolution, eliminating the race where a resume arrives before the interrupt checkpoint is durable
- **Reduction persistence hardening**: session reductions (e.g., message summarization) are persisted atomically, so a crash mid-reduction doesn't leave the session in an inconsistent state

### Model Call Timeout

`ModelTimeoutConfig` provides per-phase timeout budgets for model calls:

```go
ModelTimeoutConfig{
    CallTimeout:      10 * time.Second, // Generate/Stream opening
    FirstChunkTimeout: 5 * time.Second, // Time to first chunk
    StreamIdleTimeout: 3 * time.Second, // Inter-chunk idle
    TotalTimeout:     30 * time.Second, // Per-attempt total
}
```

The timeout wrapper sits **inside** retry and failover (per-attempt), so each attempt gets a fresh budget. `defaultIsRetryAble` only retries timeouts that occurred before any output reached downstream (`ChunksReceived == 0`), because replaying after partial output requires domain-specific logic that users can provide via custom `ShouldRetry`.

## Decisions

**Per-attempt vs per-request timeout**: The wrapper lives inside the retry/failover chain, so each attempt independently times out. A per-request timeout would require coordinating remaining budget across attempts—complexity that doesn't match the common use case of "retry if the model hung before producing anything."

**Default retry only before output**: Once chunks have been delivered downstream, the caller may have already acted on partial content. Automatic retry would produce duplicate/conflicting output. Users who handle this (e.g., by buffering) can override via `ModelRetryConfig.ShouldRetry` inspecting `ModelTimeoutError.ChunksReceived`.

**Implementation-local config (no global option functions)**: `ModelTimeoutConfig` is a plain struct field on the agent config, consistent with the project convention that per-feature configs are not registered as global options.

## Key Insight

Timeout placement relative to retry/failover determines the failure semantics:

1. Timeout **inside** retry → each attempt gets a fresh budget; a single slow attempt triggers retry while total time can still grow
2. Timeout **outside** retry → bounds total wall-clock time but doesn't distinguish "model is slow" from "model failed and retry is making progress"

For model calls, (1) is correct because the failure mode we want to detect is "this specific attempt hung"—the user already controls total wall-clock via context cancellation.

## Summary

| Problem | Solution |
|---------|----------|
| No built-in session persistence across turns | `SessionService[M]` with append-only event log and cursor-based loading |
| Race between interrupt checkpoint and resume | Defer interrupt checkpoint until session events are durable; managed resume coordination |
| Crash mid-reduction leaves inconsistent state | Atomic reduction persistence with rollback support |
| Hanging model calls block indefinitely | `ModelTimeoutConfig` with 4 independent per-phase budgets |
| Retrying after partial output causes duplication | Default retry only when `ChunksReceived == 0`; user-overridable |

---

# Runner 管理的会话持久化

## Problem

ADK Runner 没有内置的跨轮次会话状态持久化机制。需要多轮持久化的用户必须在 Runner 外部自行实现 checkpoint 管理，处理序列化、中断/恢复协调和会话生命周期。这容易出错——尤其是在管理中断恢复的竞态条件、部分输出重放和 reduction 持久化方面。

此外，模型调用没有超时控制。一个挂起的 LLM 连接（首 token 慢、流式卡住或无界总时长）会无限期阻塞，无法检测或恢复。

### 会话持久化

类型化的 `SessionService[M]` 在 Runner 内管理会话状态的完整生命周期。它使用 append-only event log（`SessionStore`）作为持久化原语，支持基于 cursor 的增量加载和带重试退避的可配置 page size。

核心能力：
- **会话回滚**：当某轮产生不良结果时回退到先前 checkpoint
- **托管中断/恢复**：Runner 协调中断 checkpoint 和恢复解析，消除 resume 在中断 checkpoint 持久化之前到达的竞态
- **Reduction 持久化加固**：会话 reduction（如消息摘要）原子持久化，崩溃不会留下不一致状态

### 模型调用超时

`ModelTimeoutConfig` 为模型调用提供分阶段超时预算：

```go
ModelTimeoutConfig{
    CallTimeout:      10 * time.Second, // Generate/Stream 发起
    FirstChunkTimeout: 5 * time.Second, // 首 chunk 等待
    StreamIdleTimeout: 3 * time.Second, // chunk 间空闲
    TotalTimeout:     30 * time.Second, // 单次尝试总时长
}
```

超时 wrapper 位于 retry 和 failover **内部**（per-attempt），每次尝试获得独立预算。`defaultIsRetryAble` 仅在输出到达下游之前（`ChunksReceived == 0`）重试超时，因为部分输出后的重放需要用户通过自定义 `ShouldRetry` 提供领域逻辑。

## Decisions

**Per-attempt vs per-request 超时**：wrapper 位于 retry/failover 链内部，每次尝试独立超时。Per-request 超时需要跨尝试协调剩余预算——这种复杂性不匹配"模型在产生输出前挂起则重试"的常见场景。

**默认仅在输出前重试**：chunk 已交付下游后，调用方可能已基于部分内容采取行动。自动重试会产生重复/冲突输出。处理此场景的用户（如通过缓冲）可通过 `ModelRetryConfig.ShouldRetry` 检查 `ModelTimeoutError.ChunksReceived` 来覆盖。

**实现本地配置（无全局 option 函数）**：`ModelTimeoutConfig` 是 agent config 上的普通结构体字段，遵循项目惯例：per-feature 配置不注册为全局 options。

## Key Insight

超时相对于 retry/failover 的位置决定了故障语义：

1. 超时在 retry **内部** → 每次尝试获得独立预算；单次慢尝试触发重试，总时间仍可增长
2. 超时在 retry **外部** → 限定总壁钟时间，但无法区分"模型慢"和"模型失败但重试在推进"

对于模型调用，(1) 是正确的，因为我们要检测的故障模式是"这次特定尝试挂起"——用户已通过 context cancellation 控制总壁钟时间。

## Summary

| Problem | Solution |
|---------|----------|
| 无内置跨轮次会话持久化 | `SessionService[M]` 配合 append-only event log 和基于 cursor 的加载 |
| 中断 checkpoint 与 resume 之间的竞态 | 延迟中断 checkpoint 至会话事件持久后；托管 resume 协调 |
| Reduction 中途崩溃导致不一致状态 | 原子 reduction 持久化 + 回滚支持 |
| 模型调用挂起无限阻塞 | `ModelTimeoutConfig` 提供 4 个独立分阶段超时预算 |
| 部分输出后重试导致重复 | 默认仅 `ChunksReceived == 0` 时重试；用户可覆盖 |


carltonawong:
One invariant I'd make explicit in the Runner session tests/docs: session history persistence and checkpoint resume are different authorities. When a run resumes, it helps to expose a small resume receipt with the session id, checkpoint id if present, last persisted turn boundary, interrupt/stop reason, and whether message history came from SessionStore or caller input. That makes it testable that durable conversation state was restored without silently treating a prior runtime boundary as current.

mrh997 (approved):

mrh997 (approved):

mrh997 (approved):

PR #1090, opened 2026-06-19
fix(adk): preserve streaming tool output during reduction

### LINKED ISSUE #1064: reduction middleware `WrapStreamableToolCall` blocks until stream is fully consumed, defeating streaming
**Describe the bug**

`typedToolReductionMiddleware.WrapStreamableToolCall` (in `adk/middlewares/reduction/reduction.go:466`) synchronously consumes the entire tool output stream inside `TruncHandler` before returning the result to the caller. This effectively turns a streaming tool call into a blocking/invoke-style call - the upstream `eventSenderToolWrapper` cannot emit incremental `AgentEvent`s until the stream is fully drained, so the end user receives all chunks at once after the tool finishes instead of in real time.

**To Reproduce**

1. Register a tool via `filesystem.StreamingShell` / `InferStreamTool` that produces output incrementally (e.g., the built-in `execute` tool backed by `ExecuteStreaming`).
2. Enable the `reduction` middleware with default config (no `SkipTruncation` for the `execute` tool).
3. Run the agent with `EnableStreaming: true`.
4. Observe that `AgentEvent` with `Role == schema.Tool` for the `execute` tool is only received **after** the command completes, with all output chunks arriving at once.

**Root cause**

In `WrapStreamableToolCall` (line 466-515):

```go
func (t *typedToolReductionMiddleware[M]) WrapStreamableToolCall(...) {
    return func(...) (*schema.StreamReader[string], error) {
        output, err := endpoint(ctx, argumentsInJSON, opts...)  // returns StreamReader
        
        readers := output.Copy(2)
        // ...
        detail := &ToolDetail{
            StreamToolResult: schema.StreamReaderWithConvert(readers[0], ...),
        }
        truncResult, err := cfg.TruncHandler(ctx, detail)  // - BLOCKS here
        // TruncHandler reads StreamToolResult to decide whether to truncate,
        // which drains the entire stream synchronously.
        // ...
        return readers[1], nil  // only returns after stream is fully consumed
    }, nil
}
```

`TruncHandler` is a synchronous function that receives a `StreamReader` via `detail.StreamToolResult`. To determine whether truncation is needed, it must consume the stream - which blocks until the underlying tool finishes producing all output. The outer `eventSenderToolWrapper.WrapStreamableToolCall` calls this as `endpoint(...)` and only proceeds to `result.Copy(2)` + `execCtx.send(event)` after it returns, so the event carrying the tool's stream is never sent until the tool is done.

**Expected behavior**

`WrapStreamableToolCall` should return a wrapped `StreamReader` immediately (without blocking), performing truncation logic incrementally as chunks are received. For example:

- Return a `StreamReader` wrapper that tracks cumulative output length during `Recv()`.
- When the cumulative length exceeds `MaxLengthForTrunc`, append a truncation notice and signal EOF.
- Handle offloading asynchronously or after the stream completes.

This would preserve true streaming semantics - the upstream `eventSenderToolWrapper` can emit each chunk as an `AgentEvent` in real time.

**Workaround**

Add `SkipTruncation: true` for streaming tools (e.g. `execute`) in the reduction config:

```go
reduction.New(ctx, &reduction.Config{
    ToolConfig: map[string]*reduction.ToolReductionConfig{
        "execute": {SkipClear: true, SkipTruncation: true},
    },
})
```

This bypasses the blocking path but also disables truncation for those tools.

**Version:**

`github.com/cloudwego/eino v0.9.2`

**Environment:**

N/A (not environment-specific, reproducible from code inspection)

**Additional context**

The `WrapInvokableToolCall` implementation (line 419) does not have this problem because invoke-style tools are inherently synchronous. The issue is specific to the streaming path where the synchronous `TruncHandler` signature is incompatible with streaming semantics.

Zhang-986:
## Summary
- use a default streaming-specific truncation notice with no tail preview, so default streaming truncation does not wait for EOF before notifying truncation
- forward chunks only until `MaxLengthForTrunc`, emit the notice once the threshold is crossed, then drain remaining output for offload without forwarding it
- keep non-streaming and tool-specific/custom `TruncHandler` behavior unchanged
- cover streamable/enhanced streamable nonblocking behavior plus default streaming terminal, error, drain, and consumer-close paths

Fixes #1064.

## Tests
- `go test ./adk/middlewares/reduction -run 'TestReductionMiddlewareTrunc/.*default.*streamable.*(terminal cases|nonblocking)' -count=1`
- `go test ./adk/middlewares/reduction -count=1`
- `go test ./adk/middlewares/reduction -coverprofile=/tmp/eino-reduction-after2.cover -covermode=count -count=1`
- local patch coverage estimate for `adk/middlewares/reduction/reduction.go`: 217/227 coverable added lines hit, about 95.59%
- wrapper coverage: `wrapDefaultStreamableTruncation` 94.4%, `wrapDefaultEnhancedStreamableTruncation` 95.5%
- `go test ./adk/... -count=1`
- `git diff --check`

Note: the latest fork workflow runs are currently waiting for maintainer approval (`action_required`), so Codecov has not recalculated for commit `95ac44d` yet.

### DIFF
--- a/adk/middlewares/reduction/consts.go
+++ b/adk/middlewares/reduction/consts.go
@@ -41,6 +41,15 @@ Preview (last {preview_size}):
 )
 
 const (
+	streamTruncFmt = `<persisted-output>
+Output truncated after {preview_size} bytes were streamed. Full output saved to: {file_path}
+</persisted-output>`
+	streamTruncFmtZh = `<persisted-output>
+输出结果在流式传输 {preview_size} 字节后被截断。完整输出保存到: {file_path}
+</persisted-output>`
+)
+
+const (
 	clearWithOffloadingFmt = `<persisted-output>Tool result saved to: {file_path}
 Use {read_tool_name} to view</persisted-output>`
 	clearWithOffloadingFmtZh = `<persisted-output>工具结果已保存至: {file_path}
@@ -58,6 +67,13 @@ func getTruncFmt() string {
 	return internal.SelectPrompt(internal.I18nPrompts{
 		English: truncFmt,
 		Chinese: truncFmtZh,
+	})
+}
+
+func getStreamTruncFmt() string {
+	return internal.SelectPrompt(internal.I18nPrompts{
+		English: streamTruncFmt,
+		Chinese: streamTruncFmtZh,
 	})
 }
 
--- a/adk/middlewares/reduction/consts.go
+++ b/adk/middlewares/reduction/consts.go
@@ -17,7 +17,11 @@
 // Package reduction provides middlewares to trim context and clear tool results.
 package reduction
 
-import "github.com/cloudwego/eino/adk/internal"
+import (
+	"fmt"
+
+	"github.com/cloudwego/eino/adk/internal"
+)
 
 const (
 	truncFmt = `<persisted-output>
@@ -42,10 +46,10 @@ Preview (last {preview_size}):
 
 const (
 	streamTruncFmt = `<persisted-output>
-Output truncated after {preview_size} bytes were streamed. Full output saved to: {file_path}
+Output truncated after {preview_size} bytes were streamed. {offload_notify} {error_msg_notify}
 </persisted-output>`
 	streamTruncFmtZh = `<persisted-output>
-输出结果在流式传输 {preview_size} 字节后被截断。完整输出保存到: {file_path}
+输出结果在流式传输 {preview_size} 字节后被截断。{offload_notify} {error_msg_notify}
 </persisted-output>`
 )
 
@@ -75,6 +79,23 @@ func getStreamTruncFmt() string {
 		English: streamTruncFmt,
 		Chinese: streamTruncFmtZh,
 	})
+}
+
+func formatStreamOffloadSavedNotify(filePath string) string {
+	return fmt.Sprintf(internal.SelectPrompt(internal.I18nPrompts{
+		English: "Full output saved to: %s.",
+		Chinese: "完整输出保存到: %s。",
+	}), filePath)
+}
+
+func formatStreamOffloadFailedNotify(err error) string {
+	if err == nil {
+		return ""
+	}
+	return fmt.Sprintf(internal.SelectPrompt(internal.I18nPrompts{
+		English: "Failed to save full output: %v.",
+		Chinese: "完整输出保存失败: %v。",
+	}), err)
 }
 
 func getClearWithOffloadingFmt() string {
--- a/adk/middlewares/reduction/reduction.go
+++ b/adk/middlewares/reduction/reduction.go
@@ -533,95 +533,135 @@ func (t *typedToolReductionMiddleware[M]) wrapDefaultStreamableTruncation(
 	argumentsInJSON string,
 	output *schema.StreamReader[string],
 ) *schema.StreamReader[string] {
-	sr, sw := schema.Pipe[string](1)
-	go func() {
-		defer sw.Close()
-		defer output.Close()
+	var (
+		fullOutput             strings.Builder
+		curSize                int
+		reachedLimit           bool
+		truncTextPrefix        string
+		truncTriggeredFullText string
+		sourceErr              error
+	)
 
-		var fullOutput strings.Builder
-		truncating := false
-		filePath := ""
-		for {
-			chunk, recvErr := output.Recv()
-			if recvErr != nil {
-				if recvErr != io.EOF {
-					var zero string
-					sw.Send(zero, recvErr)
-					return
-				}
-				break
-			}
-
-			fullOutput.WriteString(chunk)
-			if truncating {
-				continue
-			}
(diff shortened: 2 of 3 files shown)


Zhang-986:
Updated the PR to address the Codecov failures by adding targeted coverage for the new default streaming truncation paths.

What changed:
- Added terminal-path tests for default streamable truncation: short/no truncation, source stream error, trunc handler error, no-trunc decision, offload error, nil trunc result, and non-text result parts.
- Added the same terminal-path coverage for enhanced streamable truncation, including concat-error handling.
- Kept production code unchanged.

Local checks:
- `go test ./adk/middlewares/reduction -run 'TestReductionMiddlewareTrunc/.*default.*truncation|TestReductionMiddlewareTrunc/.*nonblocking' -count=1`
- `go test ./adk/middlewares/reduction -coverprofile=/tmp/eino-reduction.cover -covermode=count -count=1`
- `go test ./adk/middlewares/reduction -count=1`
- `go test ./adk/... -count=1`
- `git diff --check`

Local diff coverage estimate for `adk/middlewares/reduction/reduction.go`: 140/148 coverable added lines hit, about 94.6%.

N3kox:
@Zhang-986 hi, thanks for your tackling #1064 - the blocking behavior is real and worth fixing

However, my concern is that this PR seems to address the symptom rather than the root cause. As long as the default truncation format is `preview_first + notice + preview_last`, the streaming path will inherently need EOF to construct `preview_last`. In that sense, draining the stream is a consequence of the current prompt shape, not just an implementation detail.

That’s why I’m hesitant about this approach: it streams the prefix live, but once the threshold is crossed it still waits for EOF and then emits the full truncation notice, which includes `preview_first`. So the consumer effectively gets the head content twice, and the final persisted/history content diverges from the invoke / non-streaming path for the same tool.

I wonder if the cleaner fix is at the prompt layer: use a streaming-specific truncation format with a full-length head preview plus a trailing notice, and no tail preview. Then truncation can remain fully incremental: stream until the threshold, stop, emit the notice, and avoid both EOF waiting and duplicated content. Such as:

```
[chunk_1]
[chunk_2]
...
[chunk_n]                 ← threshold reached; stop forwarding here
                          (chunk_n+1 … chunk_m: collected & written to file, never sent)
[trunc_notify_chunk("output truncated: {original_size} total, first {preview_size} shown above. Full output saved to: {file_path}")]
EOF
```

Happy to discuss if I’m missing a constraint, but I’d be more comfortable with a prompt-level fix here.

Zhang-986:
Implemented the prompt-level/default streaming fix:

- default streamable and enhanced streamable truncation now use a streaming-specific notice with no tail preview, so the notice can be emitted before source EOF
- chunks are forwarded only up to `MaxLengthForTrunc`; after that the wrapper emits the notice, drains the remaining source output, and writes the full content to the backend without forwarding the tail
- non-streaming truncation and tool-specific/custom `TruncHandler` behavior are unchanged

Verification:
- `go test ./adk/middlewares/reduction -run 'TestReductionMiddlewareTrunc/.*(nonblocking|terminal cases)' -count=1`
- `go test ./adk/middlewares/reduction -count=1`
- `go test ./adk/middlewares/reduction -coverprofile=/tmp/eino-reduction.cover -covermode=count -count=1`
- local patch coverage estimate: 85.84% (200/233 changed statements)
- `go test ./adk/... -count=1`
- `git diff --check`

Zhang-986:
Updated the branch with commit `95ac44d` to cover the remaining meaningful default streaming truncation terminal paths.

What changed in the latest commit:
- added coverage for draining tail chunks after the truncation notice
- added backend write-error forwarding for streamable and enhanced streamable wrappers
- added enhanced concat-error coverage before notice emission
- added consumer-close early-return coverage for short, prefix, and notice sends

Local verification:
- `go test ./adk/middlewares/reduction -count=1`
- `go test ./adk/... -count=1`
- `git diff --check`
- local patch coverage estimate for `adk/middlewares/reduction/reduction.go`: 217/227 coverable added lines hit, about 95.59%

The latest fork workflow runs are currently waiting for maintainer approval (`action_required`), so Codecov has not recalculated this commit yet.

N3kox (approved):

N3kox:
@Zhang-986 
Hi, thanks again for your contribution. I made some improvements based on your commits, please take a look at it if you got time.

- Refactored default streamable truncation to use `StreamReaderWithConvert` handlers instead of an extra goroutine + pipe wrapper. And applied the same flow to enhanced streamable tool results.
- Preserved the source-error fallback behavior: if a non-EOF source error arrives after truncation has started, suppressed original chunks are replayed first, then the original source error is returned.
- Moved final truncation notice/offload work to source EOF, so the wrapper can keep accumulating the full output before offloading.
- Improved boundary handling for offload/path/backend/write failures: these are now rendered in the truncation notice instead of interrupting the stream as errors.
- Updated stream truncation prompt formatting and tests.

hi-pender (approved):

PR #1106, opened 2026-06-24
refactor(adk): introduce SessionEventVariant and simplify session event model

shentongmartin:
## 问题 (Problem)

The session event model had accumulated structural debt:

1. **Duplicate identity fields**: `TypedAgentEvent` carried `EventID` and `Timestamp` at the top level, while `SessionEvent` also had them — creating two sources of truth and confusion about which was canonical.
2. **SessionID in durable payloads**: `SessionEvent.SessionID` was part of the persisted JSON, but it is runtime ownership metadata, not event content. Stores should receive session identity as a first-class argument, not embedded in each event.
3. **Streaming message identity gap**: Streaming messages (`MessageStream`) carried no durable identity before materialization, making it impossible to correlate live stream chunks with persisted incomplete-stream events.
4. **Overloaded interrupt types**: `UserObservationEvent`, `UserInterruptEvent`, `AgentInterruptEvent`, and `AgentInterruptContext` had nested and confusing naming that didn't map cleanly to the two distinct concepts (cancel vs. business interrupt).
5. **Verbose store interface**: `AppendSessionEventsRequest` and `LoadSessionEventsRequest.SessionID` added unnecessary ceremony around simple parameters.

## 解决方案 (Solution)

Introduce `SessionEventVariant` as a sum type that cleanly separates the two live-event shapes:

```go
type SessionEventVariant[M MessageType] struct {
    SessionID string                    // live-only ownership metadata
    Event            *SessionEvent[M]   // fully materialized durable event
    MessageStreamRef *MessageStreamRef  // durable identity for a streaming message
}
```

**Key changes:**

- **`SessionEventVariant` replaces `SessionEvent` pointer in `TypedAgentEvent`**: All session-related metadata flows through one typed envelope. `SessionID` lives only on the variant (live metadata), not on `SessionEvent` (durable payload).
- **`MessageStreamRef` carries durable identity for streaming messages**: Before a stream completes or fails, its event identity (`EventID`, `Timestamp`, `Kind`, `TurnID`) is reserved and carried in the variant — enabling SSE resume correlation and incomplete-stream persistence with stable IDs.
- **Simplified `SessionEventStore` interface**: `sessionID` is passed as a first-class argument to both `LoadEvents` and `AppendEvents`. `AppendSessionEventsRequest` is removed; events are passed directly as a slice.
- **Renamed interrupt/cancel types**: `UserObservationEvent` / `UserInterruptEvent` → `CancelEvent`; `AgentInterruptEvent` / `AgentInterruptContext` → `InterruptEvent` / `InterruptContext`. Names now directly match the two concepts.
- **Removed `EventID` and `Timestamp` from `TypedAgentEvent`**: These fields belonged to the session event layer, not the agent event layer. They are accessed via `SessionEventVariant.Event` or `SessionEventVariant.MessageStreamRef`.
- **Removed `SessionRunStateRescheduled`**: Only `running` and `idle` states remain; the rescheduled state was unused.

## 关键洞察 (Key Insights)

1. **Sum types prevent invalid states**: Before, you had to check `event.SessionEvent != nil` AND separately reason about streaming vs. materialized. Now, `SessionEventVariant` makes the two states explicit — either there's a materialized `Event`, or there's a `MessageStreamRef` for an in-flight stream. No half-initialized combinations possible.

2. **SessionID is ownership, not content**: A session event's identity is `(session_id, event_id)`, but `session_id` is the container identity, not part of the event's content. Stores already know which session they're operating on; embedding it in each event payload is redundant and couples the serialization format to deployment topology.

3. **Streaming messages need identity before completion**: For partial-stream persistence and SSE resume to work correctly, a streaming message must have its durable EventID allocated when the stream starts, not when it completes. `MessageStreamRef` reserves that identity upfront.

4. **Naming clarity reduces cognitive load**: Four types (`UserObservationEvent`, `UserInterruptEvent`, `AgentInterruptEvent`, `AgentInterruptContext`) for two concepts (cancel, interrupt) created constant confusion. Two types with direct names (`CancelEvent`, `InterruptEvent`) make the code self-documenting.

---

## Problem

The session event model had accumulated structural debt with duplicate identity fields, session-id embedded in durable payloads, no durable identity for streaming messages before materialization, overloaded interrupt type naming, and verbose store interfaces.

## Solution

Introduce `SessionEventVariant` as a typed sum type carrying either a fully materialized `SessionEvent` or a `MessageStreamRef` (reserved durable identity for in-flight streams), with `SessionID` as live-only metadata on the variant. Simplify `SessionEventStore` to take `sessionID` as a first-class argument. Rename interrupt types to `CancelEvent` / `InterruptEvent` / `InterruptContext`. Remove `EventID` and `Timestamp` from `TypedAgentEvent` (they belong on the session layer).

## Key Insights

- **Sum types prevent invalid states**: `SessionEventVariant` makes materialized vs streaming explicit — no half-initialized combinations.
- **SessionID is ownership metadata, not event content**: Belongs on the live envelope, not the durable payload.
- **Streams need identity before completion**: `MessageStreamRef` reserves EventID at stream start for partial-persistence and SSE resume.
- **Naming clarity pays off**: Four types for two concepts → two types with direct names.

### DIFF
--- a/adk/chatmodel.go
+++ b/adk/chatmodel.go
@@ -375,6 +375,31 @@ func sameSystemMessage[M MessageType](oldSys, newSys M) bool {
 	}
 }
 
+func deepCopyMessage[M MessageType](msg M) M {
+	switch v := any(msg).(type) {
+	case *schema.Message:
+		cp := *v
+		if v.Extra != nil {
+			cp.Extra = make(map[string]any, len(v.Extra))
+			for k, val := range v.Extra {
+				cp.Extra[k] = val
+			}
+		}
+		return any(&cp).(M)
+	case *schema.AgenticMessage:
+		cp := *v
+		if v.Extra != nil {
+			cp.Extra = make(map[string]any, len(v.Extra))
+			for k, val := range v.Extra {
+				cp.Extra[k] = val
+			}
+		}
+		return any(&cp).(M)
+	default:
+		return msg
+	}
+}
+
 func setMessageIDFromTarget[M MessageType](msg M, targetID string) {
 	if targetID == "" || isNilMessage(msg) {
 		return
@@ -385,6 +410,8 @@ func setMessageIDFromTarget[M MessageType](msg M, targetID string) {
 func syncLeadingSystemMessageSessionEvent[M MessageType](
 	ctx context.Context,
 	previous []M,
+	oldSys M,
+	hasOldSys bool,
 	generated []M,
 ) error {
 	execCtx := getTypedChatModelAgentExecCtx[M](ctx)
@@ -398,7 +425,7 @@ func syncLeadingSystemMessageSessionEvent[M MessageType](
 	}
 
 	var event *TypedAgentEvent[M]
-	if oldSys, hasOldSys := leadingSystemMessage(previous); hasOldSys {
+	if hasOldSys {
 		EnsureMessageID(oldSys)
 		oldID := GetMessageID(oldSys)
 		setMessageIDFromTarget(newSys, oldID)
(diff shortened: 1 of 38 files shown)


adk/chatmodel.go
```diff
@@ -379,45 +359,6 @@ func leadingSystemMessage[M MessageType](messages []M) (M, bool) {
 	return zero, false
 }
 
-type leadingSystemSyncInput[M MessageType] struct {
```
shentongmartin, line 382: why do these types and functions deleted?

shentongmartin: Good catch! You found a real regression.

The original deletion of `snapshotLeadingSystemForSync` introduced a bug: when `GenModelInput` mutates the leading system message in-place (e.g. modifying `Messages[0].Extra`), the `sameSystemMessage(oldSys, newSys)` comparison would see the already-mutated "old" message, silently suppressing the `message_updated` event and persisting a stale system message.

This was fixed in commit 351f8ddd ("fix(adk): snapshot leading system message before genModelInput"):

- Added `deepCopyMessage[M]` helper (struct copy + Extra map deep copy)
- At all 3 call sites (no-tools, react, agentic-react), the leading system message is now snapshotted **before** `genModelInput` runs
- The snapshot is passed to `syncLeadingSystemMessageSessionEvent` for the comparison
- Added 2 regression tests (Extra mutation + Content mutation in GenModelInput)

The new approach is simpler than the old `leadingSystemSyncInput` + `snapshotLeadingSystemForSync` combo — just snapshot before the call and pass it as a parameter.

PR #1126, opened 2026-06-30
fix(adk): isolate agent tool interrupt state

hi-pender:
## Summary
- isolate agent-tool interrupt state by addressing nested tool interrupts with stable child paths
- preserve permission interrupt metadata across agent tool execution
- cover nested agent-tool permission/resume behavior with regression tests

## Tests
- go test ./adk/...

### DIFF
--- a/adk/agent_tool.go
+++ b/adk/agent_tool.go
@@ -19,6 +19,7 @@ package adk
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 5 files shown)


hi-pender:
Superseded by a managed-agents-side fix: the permission middleware now appends its own address segment, so this agent_tool-specific address change is no longer needed. The broader wrapper/endpoint addressing model can be revisited in a future Eino design.

PR #1127, opened 2026-07-01
fix(adk): isolate tool middleware interrupt address

shentongmartin:
# Isolate tool middleware interrupt address

## Problem
`ChatModelAgent` tool middleware and the wrapped business tool used the same execution address. When middleware interrupted a tool call, the resumed inner `agentTool` could read middleware-owned interrupt state as if it belonged to the tool, producing resume/type ownership confusion.

Example:

```text
tool:<toolName>:<callID> owns both middleware gate and inner agent tool
resume data for the middleware is visible to the inner agent tool
```

## Solution
Named tool middleware now gets its own address frame under the tool call:

```text
tool:<toolName>:<callID> -> middleware:<name>
```

The middleware runs inside that frame. When it invokes the wrapped business tool, the context is popped back to the parent tool address, so the inner tool only sees state addressed to itself. Unnamed middleware keeps the old address behavior.

## Decision
The middleware segment does not repeat `callID`. The parent tool segment already includes `callID`, so the full hierarchical address is unique without encoding the same identity twice:

```text
tool:<toolName>:<callID>;middleware:<name>
```

## Breaking API Change
This PR also removes two ADK middleware packages:

- `adk/middlewares/permission`
- `adk/middlewares/modeltimeout`

Users who still need those behaviors can copy the previous middleware implementation into their own projects or provide equivalent external middleware. Persisted sessions containing removed middleware-specific registered payloads such as `_eino_adk_permission_*` or `_eino_adk_model_timeout_error` may no longer deserialize after the packages are removed.

Core ADK compatibility hooks remain in place: `ModelTimeoutMeta` is still part of session span metadata, and timeout-aware model errors can still populate it by implementing `ModelTimeoutSpanMeta()`.

## Key Insight
Interrupt resume data is owned by an execution address. Middleware that can interrupt is not the same execution owner as the tool it wraps, so named middleware needs a distinct address frame while still calling the inner tool from the original tool address.

## Summary
| Problem | Solution |
|---------|----------|
| Middleware resume state can leak into the inner `agentTool` | Give named middleware its own `middleware:<name>` address frame under `tool:<toolName>:<callID>` |
| Inner tool execution may inherit wrapper-owned state | Pop back to the parent address before invoking the wrapped tool |
| Built-in permission and model timeout middlewares are no longer part of Eino | Remove those public packages and keep core extension hooks for external equivalents |

---

# 隔离 tool middleware 的 interrupt address

## 问题
`ChatModelAgent` 的 tool middleware 和被包裹的业务 tool 复用了同一个 execution address。当 middleware 中断一次 tool call 后，恢复执行时内部的 `agentTool` 可能把 middleware 拥有的 interrupt state 当成自己的状态读取，导致 resume/type 归属混乱。

示例：

```text
tool:<toolName>:<callID> 同时承载 middleware gate 和内部 agent tool
middleware 的 resume data 对内部 agent tool 可见
```

## 方案
具名 tool middleware 现在会在 tool call 下获得独立 address frame：

```text
tool:<toolName>:<callID> -> middleware:<name>
```

middleware 在该 frame 内执行。当它调用被包裹的业务 tool 时，context 会先 pop 回父级 tool address，因此内部 tool 只会看到属于自己的状态。未命名 middleware 保持旧的 address 行为。

## 决策
middleware segment 不再重复 `callID`。父级 tool segment 已经包含 `callID`，完整的层级地址不需要重复编码同一个身份信息也能保持唯一：

```text
tool:<toolName>:<callID>;middleware:<name>
```

## Breaking API Change
本 PR 同时移除两个 ADK middleware package：

- `adk/middlewares/permission`
- `adk/middlewares/modeltimeout`

仍需要这些能力的用户可以把旧 middleware 实现复制到自己的项目中，或提供等价的外部 middleware。包含 `_eino_adk_permission_*` 或 `_eino_adk_model_timeout_error` 等已移除 middleware 专属注册 payload 的持久化 session，在移除 package 后可能无法反序列化。

核心 ADK 兼容 hook 保留：`ModelTimeoutMeta` 仍是 session span metadata 的一部分，外部 timeout-aware model error 仍可通过实现 `ModelTimeoutSpanMeta()` 填充该 metadata。

## Key Insight
interrupt resume data 的所有权由 execution address 决定。可中断的 middleware 和它包裹的 tool 不是同一个 execution owner，因此具名 middleware 需要独立 address frame，同时调用内部 tool 时必须回到原始 tool address。

## Summary
| 问题 | 方案 |
|------|------|
| Middleware resume state 可能泄漏给内部 `agentTool` | 在 `tool:<toolName>:<callID>` 下给具名 middleware 独立的 `middleware:<name>` address frame |
| 内部 tool 执行可能继承 wrapper 拥有的状态 | 调用被包裹 tool 前先 pop 回父级 address |
| Eino 不再内置 permission 和 model timeout middleware | 移除这些公开 package，同时保留核心 extension hook 供外部等价实现使用 |


adk/handler.go
```diff
@@ -260,6 +260,30 @@ type TypedChatModelAgentMiddleware[M MessageType] interface {
...
+//     distinct frame a middleware interrupt would collide with the wrapped
+//     tool's interrupt state at the same address, and either could then observe
+//     the other's persisted state on resume.
+//
+// The returned name must be non-empty and unique within a single tool call's
+// middleware stack. NewToolNode rejects duplicate names within the same
+// endpoint kind (see validateUniqueMiddlewareNames); returning "" is treated as
+// unnamed. Because the name becomes part of the persisted interrupt address, it
+// must remain stable across restarts for checkpoints to resume correctly.
+type ToolCallWrapperNamer interface {
```
hi-pender: 是不是更泛化一点，定义成 ChatModelAgentMiddlewareNamer

shentongmartin: done

hi-pender (approved):

PR #1128, opened 2026-07-01
fix(adk): stop forwarding backgrounded sub-agent events to the parent turn

hi-pender:
## Summary
- A managed sub-agent run that is backgrounded (explicit `run_in_background`, or auto-moved to the background at the foreground deadline) runs detached and outlives the turn that launched it. It kept forwarding the inner agent's events to the parent turn's event generator, which the turn closes on end — injecting events into a stream the user has stopped watching, and racing a send against that close.
- Introduce a lifecycle signal `backgroundtask.TaskInfo.Backgrounded` (`<-chan struct{}`) that closes when a task moves to the background: before work starts for an explicit `RunInBackground` launch, and at the auto-background transition (`detach`). Foreground-completing runs never fire it.
- Add AgentTool option `adk.WithAgentToolParentForwardUntil(done)` that bounds parent forwarding by that signal. The subagent middleware wires the two together, so forwarding stops the moment a run detaches; foreground runs are unaffected.

## Design notes
- Concerns are decoupled: **backgroundtask** owns the lifecycle signal (event-agnostic), **adk/agent_tool** owns the forwarding action + the "forward until" gate, **subagent middleware** is the only layer that wires them (one line in the work closure). No reading/overriding of the parent generator option, no interposed channel/pump/atomic.
- The gated (background-capable) path uses a non-panicking send **by definition** — it may outlive its consumer. The plain foreground AgentTool path keeps a panicking `Send`, so a send on an unexpectedly-closed generator still surfaces as the bug it is.
- Known, documented residual: under auto-background the detached forwarder's lifetime exceeds the parent turn; the gate narrows but cannot fully eliminate the race with the parent close. The gated path's non-panicking send is what makes that safe, and it does not affect the synchronous path.

## Test plan
- [ ] `go test -race ./adk/...` (passes locally)
- [ ] `golangci-lint run ./adk/...` (0 issues locally)
- backgroundtask: signal timing — explicit (closed before work), foreground (stays open), auto-background (closes at detach)
- adk/agent_tool: open gate forwards, closed gate drops, closed parent generator on gated path does not panic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/adk/agent_tool.go
+++ b/adk/agent_tool.go
@@ -168,6 +168,14 @@ func (at *typedAgentTool[M]) InvokableRun(ctx context.Context, argumentsInJSON s
 	}
 
 	gen, enableStreaming := getEmitGeneratorAndEnableStreaming[M](opts)
+	// A background-capable invocation (WithAgentToolParentForwardUntil) bounds parent
+	// forwarding by a "backgrounded" signal and uses a non-panicking send: once the
+	// run detaches it outlives the parent turn, whose generator is then closed. A
+	// plain foreground invocation sets neither, so it forwards with a panicking Send
+	// — a send on an unexpectedly-closed generator is a bug there and must surface.
+	toolOpts := tool.GetImplSpecificOptions[agentToolOptions](nil, opts...)
+	gatedForward := toolOpts.gatedForward
+	forwardParentUntil := toolOpts.forwardParentUntil
 	var ms *bridgeStore
 	var iter *AsyncIterator[*TypedAgentEvent[M]]
 	var err error
@@ -274,7 +282,16 @@ func (at *typedAgentTool[M]) InvokableRun(ctx context.Context, argumentsInJSON s
 				// loop can skip them.
 				stampAgentToolSessionEvent(event, childSessionID)
 				tmp := copyTypedAgentEvent(event)
-				gen.Send(event)
+				if gatedForward {
+					// Background-capable: forward only until the run is backgrounded,
+					// then drop (the run outlives the parent turn). trySend, not Send:
+					// after detach the parent generator may be closed.
+					if !backgrounded(forwardParentUntil) {
+						gen.trySend(event)
+					}
+				} else {
+					gen.Send(event)
+				}
 				event = tmp
 			}
 		}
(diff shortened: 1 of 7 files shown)


adk/agent_tool.go
```diff
@@ -345,16 +372,55 @@ func withAgentToolOptions(agentName string, opts []AgentRunOption) tool.Option {
...
+// turn end; forwarding after that point is both wrong (the events belong to a
+// stream the user has stopped watching) and unsafe (a send on the closed generator
+// would race). Gating on done stops forwarding at the moment the run detaches.
+//
+// Presence of this option also selects the non-panicking send used on this
+// background-capable path (a plain foreground AgentTool, which never sets it,
+// forwards with a panicking send so a send on an unexpectedly-closed generator
+// surfaces as the bug it is). A nil done is treated as "never backgrounded":
+// forwarding runs for the whole invocation.
+func WithAgentToolEventForwardUntil(done <-chan struct{}) tool.Option {
```
shentongmartin: nit: 这个 option 建议不要作为 public API 暴露出来，可以直接收进 `adk/internal/...`。

它更像 ADK 内部 middleware plumbing，而不是普通 AgentTool 调用方应该感知的配置项。现在 `WithAgentToolEventForwardUntil` 只负责“转发到什么时候”，但它必须和未导出的 `withAgentToolEventForwardTarget` 配套才有完整语义。外部用户只能看到前者，看不到后者，单独使用这个 public option 基本没有意义，也会增加不必要的 API surface。

建议把 forwarding gate option 挪到 internal 包，例如 `adk/internal/agenttoolforward`：

```go
type Options struct {
    Gated bool
    Until <-chan struct{}
}

func WithUntil(done <-chan struct{}) tool.Option {
    return tool.WrapImplSpecificOptFn(func(o *Options) {
        o.Gated = true
        o.Until = done
    })
}
```

然后 `adk/agent_tool.go` 读取这个 internal option，`adk/middlewares/subagent` import 这个 internal 包来传 `task.Backgrounded`。这样 `TaskInfo.Backgrounded` 作为通用 lifecycle signal 保持公开，但 AgentTool forwarding 的内部控制不会变成外部 API。

hi-pender: 我调整下

hi-pender: done

shentongmartin (approved):

PR #1132, opened 2026-07-03
feat(summarization): add ReusePromptCaching option

mrh997:
## Summary

The summarization middleware always rebuilds a fresh model input (`[innerSysInstruction, contextMessages..., userInstruction]`) on every trigger, which fully invalidates the main conversation's cached prompt on the provider side. This change adds a `ReusePromptCaching` option so callers can opt into reusing that cache for the summarize call instead.

## API Changes

1. **Added**: `TypedConfig.ReusePromptCaching bool` (and its `Config` alias). When enabled, the summarization input becomes `[sysMsg, userMsg1, assistantMsg1, ..., userInstruction]` — the original message history is kept byte-identical and only the user instruction is appended — and the summarize call additionally passes `model.WithTools(state.ToolInfos)` and, when set, `model.WithDeferredTools(state.DeferredToolInfos)`, so the full tool envelope matches the main conversation's and the cache is preserved. Has no effect when `GenModelInput` is set.

## Key Changes

1. Because the model still sees the original messages and tools under `ReusePromptCaching`, there is a small chance it calls a tool instead of producing a summary directly — it's recommended to also configure `Retry` or `Failover` when enabling this option.
2. `summarize`/`runFailover` now thread the state's `ToolInfos` and `DeferredToolInfos` through to the model options builder so both the primary and failover calls apply the same cache-preserving options.

### DIFF
--- a/adk/middlewares/summarization/summarization.go
+++ b/adk/middlewares/summarization/summarization.go
@@ -61,6 +61,23 @@ type TypedConfig[M adk.MessageType] struct {
 	// Optional.
 	ModelOptions []model.Option
 
+	// ReusePromptCaching indicates whether the summarization call should
+	// reuse the main conversation's cached prompt instead of rebuilding a
+	// fresh input. For example, the input becomes
+	// [sysMsg, userMsg1, assistantMsg1, ..., userInstruction] rather than
+	// [innerSysInstruction, userMsg1, assistantMsg1, ..., userInstruction].
+	// The current state's ToolInfos and DeferredToolInfos are also passed
+	// via model.WithTools/model.WithDeferredTools so the tool envelope
+	// matches the main conversation's, preserving the cache.
+	//
+	// Because the model still sees the original messages and tools, there is
+	// a small chance that it chooses to call a tool instead of producing a
+	// summary directly. It is therefore recommended to also configure Retry
+	// or Failover to handle this case.
+	//
+	// Optional. Defaults to false.
+	ReusePromptCaching bool
+
 	// TokenCounter calculates the token count for given messages and tools.
 	//
 	// Parameters:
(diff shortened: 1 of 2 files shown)


adk/middlewares/summarization/summarization.go
```diff
@@ -521,6 +537,15 @@ func (m *TypedMiddleware[M]) summarize(ctx context.Context, originalMsgs []M) (M
 	return rawSummary, modelInput, nil
 }
 
+func (m *TypedMiddleware[M]) summarizeModelOptions(toolInfos []*schema.ToolInfo) []model.Option {
+	if !m.cfg.ReusePromptCaching {
+		return m.cfg.ModelOptions
+	}
+	opts := make([]model.Option, len(m.cfg.ModelOptions), len(m.cfg.ModelOptions)+1)
+	copy(opts, m.cfg.ModelOptions)
+	return append(opts, model.WithTools(toolInfos))
```
shentongmartin: 这里还需要把 `state.DeferredToolInfos` 一起传进 summarize model options。

现在 `ReusePromptCaching` 只补了 `model.WithTools(state.ToolInfos)`，但正常 ADK 模型调用会同时从 state 派生：

```go
model.WithTools(state.ToolInfos)
model.WithDeferredTools(state.DeferredToolInfos)
```

所以当主调用启用了 deferred/server-side tools 时，summarize 请求的 tool envelope 和主请求不一致：messages 前缀虽然复用了，但模型可见的工具配置少了 `DeferredToolInfos`。这会破坏这个 option 想保证的 prompt-cache 复用前提。

建议把 `DeferredToolInfos` 从 `Summarize` 一路传到 `summarizeModelOptions`，并在非 nil 时追加：

```go
opts = append(opts, model.WithTools(toolInfos))
if deferredToolInfos != nil {
    opts = append(opts, model.WithDeferredTools(deferredToolInfos))
}
```

同时补一个 `ReusePromptCaching + DeferredToolInfos` 的回归测试。

mrh997: done

shentongmartin (approved):

shentongmartin (approved):

PR #1149, opened 2026-07-18
fix(compose): avoid racing time.After(0) against task completion on immediate graph interrupt

### LINKED ISSUE #1148: adk: CancelImmediate can emit StreamCanceledError without saving a checkpoint
### Eino version

- `v0.9.12`
- also reproduced on `v0.9.13-0.20260717103118-a737972f2048` (current `main` at the time of testing)

### Problem

With a streaming `ChatModelAgent`, `Runner` configured with a `CheckPointStore`, and `adk.WithCheckPointID`, an accepted `CancelImmediate` occasionally terminates the model stream as `StreamCanceledError` instead of producing `*adk.CancelError`. In that outcome `Runner` does not save the checkpoint, and the underlying provider stream may remain blocked.

The race is reproducible with an HTTP SSE test provider that emits one `response.output_text.delta`, flushes it, and then waits for request-context cancellation. After the first delta:

```go
_, accepted := cancel(adk.WithAgentCancelMode(adk.CancelImmediate))
```

The run sometimes returns an error equivalent to:

```text
consume Eino message stream: stream canceled
```

`CheckPointStore.Get(checkpointID)` then reports `found=false`, despite `accepted=true`. Repeating the integration test (`go test -count=1000`) reproduces this reliably.

### Expected behavior

An accepted `CancelImmediate` during a model stream should consistently surface `*adk.CancelError` only after the resumable checkpoint has been persisted. The wrapped inner stream should also be released so its transport can terminate.

### Current workaround

Our application checks for the Eino checkpoint after a user-requested cancellation. If it is missing, it cancels the provider transport and persists a separate, sealed fallback marker under another checkpoint ID. On restart, Resume detects that marker and reruns the already persisted conversation history, deleting the marker only after a successful run. This preserves recoverability but cannot resume at Eino graph precision.

A fix in Eino would let us remove this fallback path.

RajeshRajendiran:
# Avoid Racing time.After(0) Against Task Completion on Immediate Graph Interrupt

Fixes #1148

## Problem

Reported in #1148: with a streaming `ChatModelAgent`, a `Runner` configured
with a `CheckPointStore`, and `adk.WithCheckPointID`, an accepted
`CancelImmediate` occasionally terminated the model stream with
`*adk.StreamCanceledError` instead of `*adk.CancelError`. When this happened,
`Runner` did not save a checkpoint, even though the cancel was reported as
accepted (`accepted=true`).

The root cause is in `compose/graph_manager.go`, not ADK. When a graph
interrupt fires with `WithGraphInterruptTimeout(0)` (which `CancelImmediate`
always uses), `receiveWithListening` still raced the interrupted task's own
completion against a `time.After(0)` timer before treating it as canceled:

```go
timeoutCh = time.After(*timeout) // *timeout == 0
...
select {
case p := <-resultCh:
    return p.ta, p.closed, false, canceled, deadline // treated as a normal completion
case <-timeoutCh:
    return nil, false, true, canceled, deadline // treated as canceled
}
```

`time.After(0)` is not instantaneous — it still goes through the Go runtime's
timer heap. Meanwhile, ADK's `wrapStreamWithCancelMonitoring` reacts to the
same cancel signal by injecting `ErrStreamCanceled` into the model's output
stream through a multi-hop goroutine/pipe chain, which completes the
interrupted task with a plain error. Depending on scheduling, that chain
could occasionally finish before the zero-duration timer fired, so
`resultCh` won the race.

When that happened, the task manager reported the task as a normal
completion (not canceled), and `graph_run.go`'s
`resolveInterruptCompletedTasks` has no way to recognize an arbitrary task
error as a cancellation — it just wraps and returns it as a fatal graph
error, entirely skipping `handleInterrupt` (where checkpoints are saved).

A smaller, related race existed in the outer `select` as well: a task
completing and a cancel signal arriving at the same instant could
non-deterministically drop the cancel signal.

## Solution

For a zero/negative interrupt timeout — used only by immediate/no-grace
interrupts — skip the timer race entirely and treat the task as canceled
right away:

```go
if *timeout <= 0 {
    now := time.Now()
    return nil, false, true, true, &now
}
```

This guarantees the checkpoint/interrupt path always runs for
`CancelImmediate`, regardless of whether the racing task happens to complete
a few microseconds earlier or later.

Also hardened the initial `select`: when the task result wins first,
non-blockingly re-check the cancel channel so an already-pending cancel
signal isn't silently treated as "no cancel happened":

```go
select {
case p := <-resultCh:
    select {
    case timeout, ok := <-cancel:
        // cancel was already pending — honor it instead of discarding it
        ...
    default:
        return p.ta, p.closed, false, false, nil
    }
case timeout, ok := <-cancel:
    ...
}
```

Non-zero and nil timeouts (safe-point cancellation, `WithAgentCancelTimeout`)
are untouched — a task that legitimately finishes within its grace period is
still used as-is.

## Decisions

**Why fix this in `compose` instead of `adk`?** The bug is in the generic
graph-interrupt/task-manager race, not in anything ADK-specific. ADK's
stream-cancel monitoring is a legitimate consumer of `WithGraphInterrupt`;
any other caller using a zero-timeout interrupt against a task racing to
complete on its own would hit the same bug.

**Why not also cancel the model's Go `context.Context` on
`CancelImmediate`?** The issue also asks that "the wrapped inner stream
should also be released so its transport can terminate." Today, ADK's root
(non-abort-only) cancel scope never cancels the Go context passed to the
model call — only `stream.Close()` runs. Wiring `ctx` cancellation into the
root scope would require also touching `graph_run.go`'s own `ctx.Done()`
check at the top of each run-loop iteration, which currently would race
against — and could bypass — the very interrupt/checkpoint path this PR
fixes. That's a separate, riskier change; left out of this PR to keep the
fix minimal and correctly scoped to the reported bug.

**Why not make the outer-select hardening fully deterministic?** It can't
be, without a different synchronization design: an unrelated task finishing
a few nanoseconds before a concurrent cancel call is a fundamentally
unresolvable race between two independent, unsynchronized signals. The
hardening added here only closes the specific, realistic pattern this bug
produces (a cancel signal already pending when the racing task result
arrives).

## Key Insight

A `time.After(duration)` value of `0` is not a synchronization primitive —
it still schedules a timer through the runtime and is not "instant" compared
to a plain channel receive. Any code that means "don't wait at all" should
short-circuit before creating the timer, not rely on winning a race against
it.

## Test plan

- New unit tests for `receiveWithListening` in
  `compose/graph_manager_test.go`: zero-timeout vs. delayed task completion,
  nil-timeout (unlimited grace) uses the real result, positive-timeout
  within/after the grace deadline.
- New end-to-end reproduction in `adk/cancel_stream_race_test.go`: a
  streaming `ChatModelAgent` whose model emits one chunk then hangs
  (matching the SSE provider described in #1148), `CancelImmediate`d 300x in
  a loop. Fails within a few iterations on the old code (`execution already
  ended` instead of a clean `*adk.CancelError` + saved checkpoint); passes
  reliably with the fix.
- `go build ./...`, `go vet ./...`, `go test ./compose/... ./adk/... -race`
  all pass.

## Summary

| Problem | Solution |
|---|---|
| Zero-timeout graph interrupt raced a `time.After(0)` timer against the interrupted task's own completion, occasionally letting a cancellation-induced task error be treated as a normal completion | Short-circuit immediate cancellation before creating the timer |
| A task completing at the same instant a cancel signal is sent to the outer `select` could silently drop the cancel signal | Non-blockingly re-check the cancel channel when the task result wins the outer select |

---

### DIFF
--- a/adk/cancel_stream_race_test.go
+++ b/adk/cancel_stream_race_test.go
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2026 CloudWeGo Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package adk
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/cloudwego/eino/components/model"
+	"github.com/cloudwego/eino/schema"
+)
+
+// deltaThenStuckChatModel mimics an HTTP SSE provider that emits one delta,
+// flushes it, and then blocks as if waiting on request-context cancellation
+// that never actually arrives (ADK's root, non-abort-only cancel scope does
+// not cancel the Go context of the model call). It never sends or closes
+// again on its own; the test unblocks all outstanding goroutines on cleanup.
+type deltaThenStuckChatModel struct {
+	delta   *schema.Message
+	started chan struct{}
+	release chan struct{}
+}
+
+func newDeltaThenStuckChatModel(delta *schema.Message) *deltaThenStuckChatModel {
+	return &deltaThenStuckChatModel{
+		delta:   delta,
+		started: make(chan struct{}, 1),
+		release: make(chan struct{}),
+	}
+}
+
+func (m *deltaThenStuckChatModel) Generate(ctx context.Context, _ []*schema.Message, _ ...model.Option) (*schema.Message, error) {
+	return m.delta, nil
+}
+
+func (m *deltaThenStuckChatModel) Stream(_ context.Context, _ []*schema.Message, _ ...model.Option) (*schema.StreamReader[*schema.Message], error) {
+	reader, writer := schema.Pipe[*schema.Message](1)
+	go func() {
+		writer.Send(m.delta, nil)
+		select {
+		case m.started <- struct{}{}:
+		default:
+		}
+		// Simulate the provider transport hanging until the connection is
+		// forcibly torn down, which never happens for a root cancel scope.
+		<-m.release
+		writer.Close()
+	}()
+	return reader, nil
+}
+
+func (m *deltaThenStuckChatModel) BindTools(_ []*schema.ToolInfo) error { return nil }
+
+// TestWithCancel_CancelImmediate_MidStream_AlwaysCheckpoints reproduces
+// https://github.com/cloudwego/eino/issues/1148: a CancelImmediate raced
+// against a streaming ChatModel that has already emitted a chunk and is
+// blocked on the next Recv() must deterministically surface *adk.CancelError
+// (never a raw StreamCanceledError) and must always persist a checkpoint.
+//
+// Before the receiveWithListening fix in compose/graph_manager.go, this
+// occasionally failed because a zero-timeout graph interrupt raced a
+// time.After(0) timer against the task's own completion (triggered by the
+// same cancel signal via wrapStreamWithCancelMonitoring's injected
+// ErrStreamCanceled), non-deterministically skipping the checkpoint/interrupt
+// path. Run repeatedly to catch the race with reasonable confidence.
+func TestWithCancel_CancelImmediate_MidStream_AlwaysCheckpoints(t *testing.T) {
+	const iterations = 300
+
+	for i := 0; i < iterations; i++ {
+		ctx := context.Background()
+
+		blk := newDeltaThenStuckChatModel(schema.AssistantMessage("hello", nil))
+		t.Cleanup(func() { close(blk.release) })
+
+		agent, err := NewChatModelAgent(ctx, &ChatModelAgentConfig{
+			Name:        "TestAgent",
+			Description: "test",
+			Model:       blk,
+		})
+		require.NoError(t, err)
+
+		store := newCancelTestStore()
+		runner := NewRunner(ctx, RunnerConfig{
+			Agent:           agent,
+			EnableStreaming: true,
+			CheckPointStore: store,
+		})
+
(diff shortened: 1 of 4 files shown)


szybnev:
Validated this PR against our HTTP SSE integration.

- baseline `a737972`: regression fails on iteration 2
- PR `9f450ea`: passes under race
- Research Harness integration: `-race -count=1000` passes with real Eino checkpoints and no fallback markers

The provider-context teardown remains separate from this PR.

adk/cancel_stream_race_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+		select {
+		case <-blk.started:
+		case <-time.After(5 * time.Second):
+			t.Fatalf("iteration %d: model did not start streaming", i)
+		}
+
+		handle, _ := cancelFn()
+		cancelErr := handle.Wait()
+		require.NoError(t, cancelErr, "iteration %d", i)
```
shentongmartin, line 120: This assertion is still racing with normal run completion in CI. The job at `compatibility-test (1.19)` failed here with `execution already ended` on iteration 44, and `eino-unit-test` failed similarly on iteration 75. Waiting for `blk.started` only proves the fake model emitted its first chunk; it does not prove the runner has reached and remains in the graph-interrupt/checkpoint path. Under `-race`/coverage instrumentation, the run can reach `stateDone` before `cancelFn()` wins, so `handle.Wait()` can legitimately return `ErrExecutionEnded`.

Please tighten the synchronization so the test cancels only after the graph interrupt hook is definitely registered and the run is still in the intended window, or treat `ErrExecutionEnded` as the late-cancel outcome and only assert `CancelError`/checkpoint persistence when the cancel was actually accepted.

RajeshRajendiran: Good catch — turned out to be more than a test sync issue. Dug into it and there was a real second race in `sendImmediateInterrupt()`: it closed `immediateChan` (which wakes the model-stream cancel monitor) before actually firing the compose graph interrupt. So occasionally the injected `ErrStreamCanceled` would fail the node before the graph interrupt signal was even sent, and the task manager would see a plain node error instead of a cancellation — same symptom, different cause than what #1149 originally fixed.

Fixed by reordering: fire the graph interrupt first, then close `immediateChan`, for the direct/non-recursive path. Left the recursive-with-descendant path as-is since it needs `immediateChan` to close early for descendant teardown during the grace period.

Ran the repro 30,000 times (100×300) under `-race` with zero failures, plus the full `adk`/`compose` suites and 10 passes of the recursive cancel tests — all clean. Pushed as a new commit, should be good for another look.

shentongmartin (approved):

PR #1158, opened 2026-07-22
feat(schema): track cache creation tokens

### LINKED ISSUE #1150: feat(schema): track LLM cache-creation (write) tokens in PromptTokenDetails
**Is your feature request related to a problem? Please describe.**

`schema.PromptTokenDetails` models only CachedTokens (cache read). There's no field for cache write (creation) tokens. Multiple providers bill cache writes at a premium and report a distinct count - Anthropic cache_creation_input_tokens (1.25x/2x), OpenAI cache_write_tokens (newer manual-cache models, 1.25x), AWS Bedrock cacheWriteInputTokens - but eino has nowhere to put it, so bindings fold it into `PromptTokens`. The total is preserved, but the distinct write count isn't surfaced:

```go
promptTokens := int(usage.InputTokens + usage.CacheReadInputTokens + usage.CacheCreationInputTokens)
PromptTokenDetails{ CachedTokens: int(usage.CacheReadInputTokens) } // write count not surfaced
```

Result: write tokens are counted, but unattributable - they land in the input bucket and get priced at 1x instead of the provider's write rate, so cost tracking under-reports the write premium and can't break out write volume.

**Describe the solution you'd like**

```go
type PromptTokenDetails struct {
    CachedTokens        int `json:"cached_tokens"`          // cache read
    CacheCreationTokens int `json:"cache_creation_tokens"`  // cache write
}
```

**Describe alternatives you've considered**

- Compute downstream - impossible; the write count is already summed into `PromptTokens` at the binding and can't be separated back out.
- Carry via ResponseMeta.Extra - works but unshaped and per-consumer; a typed field is the right home.

**Additional context**

Every other provider-agnostic framework already models write separately and provider-neutrally: LangChain / LangChain.js (input_token_details.cache_creation), Vercel AI SDK (inputTokens.cacheWrite), LlamaIndex (cache_creation_input_tokens). Gemini is the one provider with no per-request write count (write billed in a separate caches.create call) — a neutral field simply stays zero there.

March-77:
#### What type of PR is this?

feat

#### Check the PR title.

- [x] This PR title matches the format: `<type>(optional scope): <description>`.
- [x] A separate user-documentation PR is not required; the additive Go API field is documented inline.

feat(schema): 追踪提示词缓存写入 token

en:

Providers can report and bill prompt-cache writes separately from cache reads, but Eino previously exposed only `CachedTokens`. This made cache-creation volume unavailable to provider integrations and downstream cost accounting.

This PR:

- adds provider-neutral `CacheCreationTokens` fields to `schema.PromptTokenDetails` and the model callback DTO;
- preserves cache-creation counts through both message concatenation paths using the existing cumulative-usage maximum semantics;
- adds JSON round-trip and legacy-payload compatibility coverage;
- verifies cache-read and cache-write maxima independently in standard and agentic message tests.

The change is additive. Existing payloads that omit `cache_creation_tokens` continue to decode with the Go zero value.

Validation:

- `go test ./schema ./components/model`
- `go vet ./...`
- `golangci-lint run --new-from-rev=origin/main --timeout 5m` (0 issues)
- `git diff --check`

`go test ./...` was also attempted on Windows. The changed packages pass, while unrelated existing filesystem/path and CRLF-sensitive tests fail under Windows. A WSL rerun could not start because the local WSL instance timed out before `go version`.

zh(optional):

新增独立的缓存写入 token 统计，并在普通及 Agentic 流式消息合并中保留该累计值，便于准确计算提示词缓存写入成本。

Fixes #1150

N/A

### DIFF
--- a/components/model/callback_extra.go
+++ b/components/model/callback_extra.go
@@ -44,8 +44,10 @@ type CompletionTokensDetails struct {
 
 // PromptTokenDetails provides a breakdown of prompt token usage.
 type PromptTokenDetails struct {
-	// Cached tokens present in the prompt.
+	// CachedTokens is the number of prompt tokens read from a provider cache.
 	CachedTokens int
+	// CacheCreationTokens is the number of prompt tokens written to a provider cache.
+	CacheCreationTokens int
 }
 
 // Config is the config for the model.
--- a/schema/agentic_message.go
+++ b/schema/agentic_message.go
@@ -1126,6 +1126,9 @@ func concatTokenUsage(usages []*TokenUsage) *TokenUsage {
 		if usage.PromptTokenDetails.CachedTokens > ret.PromptTokenDetails.CachedTokens {
 			ret.PromptTokenDetails.CachedTokens = usage.PromptTokenDetails.CachedTokens
 		}
+		if usage.PromptTokenDetails.CacheCreationTokens > ret.PromptTokenDetails.CacheCreationTokens {
+			ret.PromptTokenDetails.CacheCreationTokens = usage.PromptTokenDetails.CacheCreationTokens
+		}
 		if usage.CompletionTokensDetails.ReasoningTokens > ret.CompletionTokensDetails.ReasoningTokens {
 			ret.CompletionTokensDetails.ReasoningTokens = usage.CompletionTokensDetails.ReasoningTokens
 		}
--- a/schema/agentic_message_test.go
+++ b/schema/agentic_message_test.go
@@ -811,7 +811,8 @@ func TestConcatAgenticMessages(t *testing.T) {
 						PromptTokens:     10,
 						CompletionTokens: 5,
 						PromptTokenDetails: PromptTokenDetails{
-							CachedTokens: 3,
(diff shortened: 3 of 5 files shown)


March-77:
Revalidation update (2026-08-23): this remains mergeable against current main (`ebd616c`); main is only two unrelated commits ahead, and neither main nor `alpha/10` contains an equivalent `CacheCreationTokens` field. Local validation passes with `go test -race ./schema ./components/model`.

The Aug 4 scope question on #1150 is valid: this PR establishes the core schema/callback contract, while current eino-ext still stores Claude cache-creation usage in message `Extra` and its Langfuse/Cozeloop mappings do not consume a cache-write field. I suggest treating provider/observability wiring as explicit eino-ext follow-ups after this core contract lands.

There are still no GitHub Actions workflow runs on this fork PR (only the passing CLA status), so `BLOCKED` reflects required review, not a failed check. A maintainer review of the core API direction—and workflow approval if the fork banner is shown—would unblock the next step.

PR #1164, opened 2026-07-24
feat(adk): expose current messages to middleware

### LINKED ISSUE #868: Provide means to access Messages for ChatModelAgentMiddleware
Since `ProcessState` is `compose` layer API, we prefer to hide it from ChatModelAgent users. To fully replace it's functionality, we need to provide first-class support to 'Get' current message list in ChatModelAgentMiddleware, other than from Before/AfterModelRewriteState. 

We already have concrete user request for this feature.

March-77:
#### What type of PR is this?

feat

#### Check the PR title.

- [x] This PR title matches the required format.
- [x] No separate documentation PR is required; the new API has Go documentation.

feat(adk): 向中间件开放当前消息列表

en:

ChatModelAgent middleware hooks that do not receive `TypedChatModelAgentState` currently have no first-class way to inspect the current message list. Consumers must reach into `compose.ProcessState` and the deprecated concrete `adk.State`; this is impossible for `AgenticMessage` because its internal state type is intentionally unexported.

This change adds:

- `GetMessages(ctx)` for the standard `*schema.Message` path
- `TypedGetMessages[M](ctx)` for both supported message types
- a shallow snapshot contract so changing the returned slice cannot mutate agent state
- clear errors when called outside an active ChatModelAgent execution

The skill middleware now consumes the public API instead of deprecated/internal state details, which also removes its message-state access restriction for the Agentic path.

User impact: middleware and tool hooks can inspect current messages without depending on compose internals. Existing APIs and behavior are unchanged.

Compatibility/risk: additive public API only. Message objects are intentionally shared and documented as read-only; only the returned slice is copied.

Validation:

- `go test ./adk -run "TestGetMessages|TestAgenticChatModelAgentRun_WithMiddleware" -count=10`
- `go test -race ./adk -run "TestGetMessages|TestAgenticChatModelAgentRun_WithMiddleware" -count=1`
- `go test ./adk/middlewares/skill`
- `go test ./adk`
- `go test ./...`
- `go vet ./...`

`golangci-lint` could not be completed locally because the pinned runner repeatedly stalled while downloading `github.com/MirrexOne/unqueryvet v1.5.4`.

Fixes #868

### DIFF
--- a/adk/agentic_test.go
+++ b/adk/agentic_test.go
@@ -402,6 +402,12 @@ func TestAgenticChatModelAgentRun_WithMiddleware(t *testing.T) {
 
 	mw := &testAgenticMiddleware{
 		beforeFn: func(ctx context.Context, state *TypedChatModelAgentState[*schema.AgenticMessage], mc *TypedModelContext[*schema.AgenticMessage]) (context.Context, *TypedChatModelAgentState[*schema.AgenticMessage], error) {
+			currentMessages, err := TypedGetMessages[*schema.AgenticMessage](ctx)
+			require.NoError(t, err)
+			require.Equal(t, state.Messages, currentMessages)
+			currentMessages[0] = nil
+			require.NotNil(t, state.Messages[0], "returned slice must not alias state slice")
+
 			state.Messages = append(state.Messages, schema.UserAgenticMessage("extra"))
 			return ctx, state, nil
 		},
--- a/adk/handler.go
+++ b/adk/handler.go
@@ -326,6 +326,32 @@ func processTypedState(ctx context.Context, fn func(extra map[string]any) map[st
 	})
 }
 
+// TypedGetMessages returns the current ChatModelAgent message list.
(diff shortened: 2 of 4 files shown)


March-77:
Revalidation update (2026-08-23): this remains mergeable against current main (`ebd616c`), with no equivalent `GetMessages`/`TypedGetMessages` API on main or `alpha/10`. The two main commits since this PR's base touch unrelated middleware packages.

Local validation passes:
- `go test ./adk ./adk/middlewares/skill`
- `go test -race ./adk -run 'TestGetMessages|TestAgenticChatModelAgentRun_WithMiddleware' -count=1`

The PR still matches the first-class API direction requested in #868 and removes the skill middleware's Agentic fork-mode restriction. GitHub Actions have never started for this fork PR (CLA is passing), so the blocked state is required review rather than a failing check. A maintainer review/workflow approval when convenient would be appreciated.

PR #1165, opened 2026-07-24
fix(adk): bound summarization model input

### LINKED ISSUE #940: 关于adk/middlewares/summarization/summarization.go中总结时的输入可能超出模型上下文窗口的问题
需要处理默认情况下用户消息超出了总结模型上下文窗口的问题：

可以考虑默认情况下使用触发阈值作为限制：滑动窗口策略淘汰旧消息，避免用于总结的消息超过了总结模型本身的上下文限制

March-77:
#### What type of PR is this?

fix

#### Check the PR title.

- [x] This PR title matches the required format.
- [x] No separate documentation PR is required; the behavior is documented on the existing config field.

fix(adk): 限制摘要模型的输入长度

en:

When conversation history crosses the summarization threshold, the default input builder currently forwards the complete over-limit history to the summarization model. If that model has the same or a smaller context window, the recovery mechanism fails with another context-length error.

This change uses `Trigger.ContextTokens` as the input budget for the default summary request. After reserving space for the system and user summary instructions, it keeps the newest contiguous message window that fits and evicts older messages. Keeping a suffix preserves current user intent and tool-call/result ordering. Custom `GenModelInput` callbacks remain untouched and retain full control.

It also fixes a related trigger edge case: configuring only `ContextMessages` previously left `ContextTokens` at zero, causing any non-empty history below the message threshold to trigger anyway. A disabled token condition now falls back to the documented default 160k threshold.

User impact: default summarization no longer retries an already oversized history unchanged, and message-only trigger configurations behave as documented.

Compatibility/risk: no new API. The window applies only to the default model-input builder; users needing different retention semantics can continue using `GenModelInput`.

Validation:

- `go test ./adk/middlewares/summarization -run "TestMiddlewareShouldSummarize|TestMiddlewareBuildSummarizationModelInput|TestWindowMessagesByTokenLimit" -count=10`
- `go test -race ./adk/middlewares/summarization -run "TestMiddlewareShouldSummarize|TestMiddlewareBuildSummarizationModelInput|TestWindowMessagesByTokenLimit" -count=1`
- `go test ./adk/middlewares/summarization`
- `go test ./adk`
- `go test ./...`
- `go vet ./...`

`golangci-lint` could not be completed locally because the pinned runner repeatedly stalled while downloading `github.com/MirrexOne/unqueryvet v1.5.4`.

Fixes #940

### DIFF
--- a/adk/middlewares/summarization/summarization.go
+++ b/adk/middlewares/summarization/summarization.go
@@ -75,6 +75,11 @@ type TypedConfig[M adk.MessageType] struct {
 
 	// Trigger specifies the conditions that activate summarization.
 	// Optional. Defaults to triggering when total tokens exceed 160k.
+	//
+	// ContextTokens also bounds the default summarization model input. When
+	// history exceeds that limit, the default input builder keeps the newest
+	// contiguous messages that fit and evicts older messages. A custom
+	// GenModelInput retains full control and is not windowed.
 	Trigger *TriggerCondition
 
 	// EmitInternalEvents indicates whether internal events should be emitted during summarization,
@@ -372,7 +377,7 @@ func (m *TypedMiddleware[M]) shouldSummarize(ctx context.Context, input *TypedTo
 
 func (m *TypedMiddleware[M]) getTriggerContextTokens() int {
 	const defaultTriggerContextTokens = 160000
-	if m.cfg.Trigger != nil {
+	if m.cfg.Trigger != nil && m.cfg.Trigger.ContextTokens > 0 {
 		return m.cfg.Trigger.ContextTokens
 	}
 	return defaultTriggerContextTokens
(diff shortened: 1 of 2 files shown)


March-77:
Revalidation update (2026-08-23): while rechecking the new token-window logic, I found that the budget boundary could land between an assistant tool call and its result, leaving an orphaned leading tool-result message that provider APIs may reject. I pushed c957055 to drop such leading results after windowing and added regression coverage for both Message and AgenticMessage paths.

Validation:
- `go test ./adk/middlewares/summarization`
- `go test -race ./adk/middlewares/summarization -run 'TestWindowMessagesByTokenLimit|TestMiddlewareBuildSummarizationModelInput' -count=1`
- `git diff --check`

Against current main (`ebd616c`), the PR remains mergeable and no equivalent implementation exists. GitHub Actions still have no workflow runs for this fork PR (CLA is passing), so the current blocked state is required review rather than a failing check.

PR #1168, opened 2026-07-29
feat(adk): add PatchedToolResultGenerator for EnhancedTool-style results

jesse-engineer:
## Summary

- Deprecate `PatchedContentGenerator` (plain string) in `adk/middlewares/patchtoolcalls`.
- Add `PatchedToolResultGenerator`, aligned with EnhancedTool: receives `*schema.ToolArgument` and returns `*schema.ToolResult`.
- When both are set, `PatchedToolResultGenerator` takes precedence; otherwise fall back to the legacy generator, then the default cancel message.
- Message path maps `ToolResult` to tool messages via `UserInputMultiContent` (and sets `Content` for single-text results). AgenticMessage path maps to `FunctionToolResult` / tool-search result blocks.

This lets callers customize patched dangling-tool messages using the original tool-call arguments (e.g. inspect chapter write payload when a confirmation interrupt is abandoned).

## Test plan

- [x] `go test ./adk/middlewares/patchtoolcalls/ -count=1`
- [x] Existing generic Message / AgenticMessage cases still pass
- [x] New cases: result generator receives arguments; result generator overrides content generator; single-text result sets both `Content` and `UserInputMultiContent`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/adk/middlewares/patchtoolcalls/patchtoolcalls.go
+++ b/adk/middlewares/patchtoolcalls/patchtoolcalls.go
@@ -26,14 +26,25 @@ import (
 	"github.com/cloudwego/eino/schema"
 )
 
+// PatchedToolResult is the value returned by PatchedToolResultGenerator.
+//
+// Content and ToolResult are mutually exclusive:
+//   - Content: plain-text tool message (typical for InvokableTool users)
+//   - ToolResult: EnhancedTool-style result (text and/or multimodal parts)
+type PatchedToolResult struct {
+	Content    string
+	ToolResult *schema.ToolResult
+}
+
 // Config defines the configuration options for the patch tool calls middleware.
 type Config struct {
 	// PatchedContentGenerator is an optional custom function to generate the content
 	// of patched tool messages as a plain string.
 	//
-	// Deprecated: Use PatchedToolResultGenerator instead, which aligns with EnhancedTool
-	// (ToolArgument in, ToolResult out) and receives the original tool-call arguments.
-	// Kept for backward compatibility; ignored when PatchedToolResultGenerator is set.
+	// Deprecated: Use PatchedToolResultGenerator instead, which receives the original
+	// tool-call arguments and can return either plain Content or an EnhancedTool
+	// ToolResult. Kept for backward compatibility; ignored when
+	// PatchedToolResultGenerator is set.
 	//
 	// Parameters:
 	//   - ctx: the context for the operation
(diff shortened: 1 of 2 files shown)


adk/middlewares/patchtoolcalls/patchtoolcalls.go
```diff
@@ -257,6 +314,147 @@ func createPatchedAgenticToolMessage(ctx context.Context, gen func(ctx context.C
...
+					CallID: callID,
+					Name:   toolName,
+					Result: result.Parts[0].ToolSearchResult,
+				}),
+			},
+		}
+	}
+
+	blocks := toolResultToFunctionBlocks(result)
+	if len(blocks) == 0 {
```
shentongmartin, line 411: 这里需要把 `toolResultToAgenticMessage` / `toolResultToFunctionBlocks` 改成返回 error，而不是在 `blocks` 为空时直接补空文本。

现在如果 `PatchedToolResultGenerator` 返回 `ToolOutputPart{Type: ToolPartTypeImage, Image: nil}`，Message 路径会在 `ToMessageInputParts()` 报错，但 Agentic 路径会因为 nil media 被跳过，最后走到这里变成一个空字符串的成功 tool result。这会把 malformed generator output 静默写进 history。

建议对 nil image/audio/video/file payload 和未知 part type 返回错误，并从 `createPatchedAgenticToolMessage` 向外传播，和 Message 路径保持一致。

jesse-engineer: fixed

adk/middlewares/patchtoolcalls/patchtoolcalls.go
```diff
@@ -40,6 +44,33 @@ type Config struct {
...
+	// Parameters:
+	//   - ctx: the context for the operation
+	//   - toolName: the name of the tool that was called
+	//   - toolCallID: the id of the tool call
+	//   - toolArgument: the original tool-call arguments (Text is the JSON arguments string)
+	//
+	// Returns:
+	//   - *schema.ToolResult: the patched tool result to insert into history
+	//   - error: any error that occurred during generation
+	PatchedToolResultGenerator func(
```
shentongmartin, line 74: suggestion: before this API is frozen, I think it is worth considering whether the replacement should stay this tightly coupled to the EnhancedTool-shaped `*schema.ToolResult` return type.

The core new capability here is argument-aware patching. For users whose tools are ordinary `InvokableTool`s and whose results are still plain `Content` strings, requiring `PatchedToolResultGenerator` means they have to wrap a string into `schema.ToolResult{Parts: ...}` just to read the original arguments. That makes the common string-result path feel like it has to opt into the EnhancedTool mental model.

One API shape that would preserve both use cases is to make the new generator return a small patch-specific result wrapper, for example:

```go
type PatchedToolResult struct {
    Content string
    ToolResult *schema.ToolResult
}
```

with `Content` and `ToolResult` mutually exclusive. Then normal `InvokableTool` users can return `Content` directly, while multimodal / EnhancedTool users can still return `ToolResult`. Another smaller option is an argument-aware string generator alongside this one. The important bit is keeping plain string tool results first-class, not just backward-compatible.

jesse-engineer: Good call — addressed in c101c5a.

`PatchedToolResultGenerator` now returns:

```go
type PatchedToolResult struct {
    Content    string
    ToolResult *schema.ToolResult
}
```

`Content` and `ToolResult` are mutually exclusive. Plain InvokableTool users can return `Content` directly; EnhancedTool/multimodal users keep using `ToolResult`.

shentongmartin (approved):
Reviewed latest head c101c5a1. Previous blocker is fixed, the wrapper API resolves the InvokableTool ergonomics concern, and focused tests pass. No remaining blockers from my side.

PR #1240, opened 2026-09-02
fix(adk): sanitize gob-unsafe error values when encoding agent event checkpoints

### LINKED ISSUE #782: 多Agent模式下其中一个SubAgent的internal错误灰导致后续其它SubAgent的CheckPont保存序列化失败
**Describe the bug**

一个supervisor agent调用SubAgentA, SubAgentA抛出innternalNodeError（比如exceed max iteration），SubAgentA transfer to MainAgent, MainAgent调用SubAgentB，SubAgentB事InterruptableAgent，获取user response后回因为SubAgentA的internal错误导致checkpoint保存gob序列化compose.InternalErr没有注册错误

**Version:**
0.7.30

lmsn-kang:
## Problem

Fixes #782.

In a supervisor multi-agent flow, an internal error from one sub-agent (e.g. exceeding max iterations) leaves a `compose.internalError` value inside the run session (events are recorded unconditionally, including failed ones). When a later sub-agent interrupts and the runner saves a checkpoint, gob encoding of the session fails:

```
failed to save checkpoint: failed to encode checkpoint:
failed to gob encode agent event wrapper:
gob: type not registered for interface: compose.internalError
```

The interrupted sub-agent is never resumable, and the user's reply to the interrupt is effectively lost. I reproduced this end-to-end (supervisor + one sub-agent exceeding `MaxIterations` + one interruptible sub-agent + in-memory `CheckPointStore`, mock models, no real LLM) on **v0.9.18** and **v0.10.0-alpha.27** — the reported v0.7.30 behavior is alive in both.

## Root cause

Errors reach the persisted session through two channels; only one was sealed:

1. **Stream channel** — an error chunk inside `MessageStream`. Already handled: `agentEventWrapper.GobEncode` materializes the stream via `consumeStream()` before encoding (guarded by `TestStreamRetryThenToolInterruptCheckpoint`).
2. **Direct channel** — `AgentEvent.Err` itself. A failed sub-agent run emits an event with `Err = *compose.internalError` (unexported, never gob-registered). `GobEncode` never looked at this field.

Simply registering the concrete type is not viable: `internalError` is unexported, and its `origError error` field can hold arbitrary user-created errors, so the registration set is unbounded.

## Fix

Encode-time sanitization at the persistence boundary (`adk/runctx.go`):

- New `gobSafeError` type: keeps only the error message, registered via `gob.RegisterName("_eino_adk_gobSafeError", ...)`.
- `agentEventWrapper.GobEncode` now builds a **detached copy** of the event with `Err` (and `StreamErr`) replaced by `gobSafeError`. The live event object — which is shared with consumers — is never mutated, and the gob wire format is unchanged (same struct type, same exported fields), so `GobDecode` needs no changes.
- `typedAgentEventWrapper[M].GobEncode` had the same hazard through `TypedAgentEvent.Err`; it gets the identical treatment on its `forGob` copy.

Semantics deliberately **not** changed: after SubAgentA fails, the flow still continues to SubAgentB (the issue thread discusses whether it should abort instead — that is a flow-semantics decision I left to the maintainers; this PR only guarantees that whatever happens, the checkpoint stays encodable). Trade-off to be aware of: after resume, restored errors are `gobSafeError`, so `errors.Is` against original sentinels no longer matches — only the message survives, which is the standard price of a persistence boundary.

## Test

- `adk/prebuilt/supervisor/issue782_checkpoint_test.go` — regression test mirroring the issue: SubAgentA exceeds `MaxIterations`, flow continues to interruptible SubAgentB; asserts the checkpoint save produces no gob error, the interrupt is delivered, and the resume read-back is clean. Fails with the reported gob error before the fix, passes after.
- Also verified end-to-end with a standalone mock-model scenario against v0.9.18 / v0.10.0-alpha.27 (broken) and this branch (green), including the resume phase.

## Note on local verification

I could not run the complete `./adk/...` suite locally (machine ran out of memory compiling the core `adk` package); `./adk/prebuilt/...` and `./compose`-dependent paths pass, and I'd appreciate CI confirming the rest.

### DIFF
--- a/adk/prebuilt/supervisor/issue782_checkpoint_test.go
+++ b/adk/prebuilt/supervisor/issue782_checkpoint_test.go
@@ -1,0 +1,219 @@
+package supervisor
+
+// Regression test for issue #782: in a supervisor multi-agent flow, an
+// internal error from one sub-agent (e.g. exceeding max iterations) used to
+// leave a gob-unencodable error value (compose.internalError) inside the run
+// session. When a later sub-agent interrupted and the runner saved a
+// checkpoint, gob failed with "type not registered for interface:
+// compose.internalError", breaking the whole interrupt/resume flow.
+//
+// With the gobSafeError sanitization in agentEventWrapper.GobEncode, the
+// checkpoint save must succeed.
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/cloudwego/eino/adk"
+	"github.com/cloudwego/eino/components/model"
+	"github.com/cloudwego/eino/components/tool"
+	"github.com/cloudwego/eino/compose"
+	"github.com/cloudwego/eino/schema"
+)
+
+type issue782Store struct {
+	mu sync.Mutex
+	m  map[string][]byte
+}
+
+func (s *issue782Store) Set(_ context.Context, key string, value []byte) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.m[key] = value
+	return nil
+}
+
+func (s *issue782Store) Get(_ context.Context, key string) ([]byte, bool, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	v, ok := s.m[key]
+	return v, ok, nil
+}
+
+// issue782Model either returns scripted messages in order, or (when
+// alwaysToolCall is set) keeps returning a tool call so that the agent hits
+// its MaxIterations limit and fails with an internal error.
+type issue782Model struct {
+	name           string
+	steps          []*schema.Message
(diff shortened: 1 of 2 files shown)


lmsn-kang:
补充说明：本 PR 只处理 checkpoint 序列化这一层（编码时消毒 gob-unsafe 的错误值），作为纵深防御建议保留：即使上游语义修复（出错后 supervisor 应中止）合入，错误事件仍会经 `parentSession.addEvent` 进入共享 session，且 interrupt 桥接等其他路径也可能把不可编码的值带进 checkpoint。

错误传播语义的根因（SubAgentA 出错后未中止、继续 transfer，参见 issue 中 @shentongmartin 的评论）我计划在 #782 补充设计讨论，确认方向后会另开 PR，与本 PR 不冲突。

PR #1267, opened 2026-09-09
fix(adk): preserve natural subagent completion during drain

N3kox:
#### What type of PR is this?

fix

#### Check the PR title.

- [x] This PR does not require user-facing documentation updates.

en:

When a drain request raced with a subagent's natural completion, the background task executor queued the drain control before cancellation completed and ignored the result of `CancelHandle.Wait`.

If the subagent completed before cancellation took effect, `CancelHandle.Wait` returned `ErrExecutionEnded`. However, the queued drain control still overrode the completed result and required a resumable checkpoint. Since natural completion does not produce a checkpoint, the executor returned `ErrDrainCheckpointUnavailable`, leaving the task running until lease recovery incorrectly marked it as failed.

This PR preserves the cancellation outcome and lets the natural completed result win when a drain loses the race with execution completion. Drains that take effect continue to require a resumable checkpoint and remain fail-closed if that checkpoint is unavailable.

A regression test verifies that the task is committed as completed, its final result is retained, and no checkpoint is required when execution completes naturally before drain cancellation takes effect.

zh(optional):

N/A

N/A

### DIFF
--- a/adk/backgroundtask/subagent/subagent.go
+++ b/adk/backgroundtask/subagent/subagent.go
@@ -262,8 +262,9 @@ func (e *Executor[M]) ValidateExecution(_ context.Context, task *backgroundtask.
 	return e.ValidateSpec(task.Spec)
 }
 
-// SupportsDrain reports true because sub-agent drain captures an ADK Runner
-// checkpoint before returning a suspended result.
+// SupportsDrain reports true because an effective sub-agent drain captures an
+// ADK Runner checkpoint and returns a suspended result. If the run terminates
+// before drain cancellation takes effect, its own outcome is preserved.
 func (e *Executor[M]) SupportsDrain() bool { return true }
 
 func validateSpecPayload(spec backgroundtask.Spec) (*taskPayload, error) {
@@ -444,26 +445,16 @@ func (e *Executor[M]) Execute(
 	})
 	cancelOption, cancelRun := adk.WithCancel()
 	controlRequests := make(chan backgroundtask.ControlRequest, 1)
+	cancelOutcomes := make(chan error, 1)
 	controlWatchDone := make(chan struct{})
 	defer close(controlWatchDone)
 	go func() {
 		select {
 		case control := <-runtime.Controls():
 			controlRequests <- control
-			cancelOptions := []adk.AgentCancelOption{adk.WithRecursive()}
-			if control.Kind == backgroundtask.ControlDrain {
-				cancelOptions = append(cancelOptions,
-					adk.WithAgentCancelMode(adk.CancelAfterChatModel|adk.CancelAfterToolCalls))
-				if e.drainCancelTimeout > 0 {
-					cancelOptions = append(cancelOptions,
-						adk.WithAgentCancelTimeout(e.drainCancelTimeout))
-				}
-			} else {
-				cancelOptions = append(cancelOptions, adk.WithAgentCancelMode(adk.CancelImmediate))
-			}
-			if handle, accepted := cancelRun(cancelOptions...); accepted {
-				_ = handle.Wait()
-			}
+			cancelOutcomes <- requestCancelAndWait(
+				control, cancelRun, e.drainCancelTimeout,
+			)
 		case <-controlWatchDone:
 		case <-ctx.Done():
 		}
@@ -485,7 +476,9 @@ func (e *Executor[M]) Execute(
 			interrupted = event.Action.Interrupted
 		}
 		if event.Err != nil && interrupted == nil {
-			return e.handleRunError(ctx, iter, task, controlRequests, event.Err)
+			return e.handleRunError(
+				ctx, iter, task, controlRequests, cancelOutcomes, event.Err,
+			)
 		}
 		materialized := event
 		if event.Output != nil && event.Output.MessageOutput != nil {
@@ -495,7 +488,9 @@ func (e *Executor[M]) Execute(
 				if errors.As(messageErr, &retryErr) {
 					continue
 				}
-				return e.handleRunError(ctx, iter, task, controlRequests, messageErr)
+				return e.handleRunError(
+					ctx, iter, task, controlRequests, cancelOutcomes, messageErr,
+				)
 			}
 			materialized = materializedEvent(event, message)
 			final = agenttool.ExtractTextContent(message)
@@ -504,15 +499,49 @@ func (e *Executor[M]) Execute(
 			materialized, foregroundcoord.ProjectionDetached(ctx), copyMaterializedEvent[M],
 		)
 	}
-	if controlResult, controlErr, controlled := e.controlResult(ctx, task, pollControl(controlRequests)); controlled {
-		return controlResult, controlErr
+	control := pollControl(controlRequests)
+	cancelOutcome := waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+	return e.resolveRunOutcome(ctx, task, control, cancelOutcome, subagentRunOutcome{
+		final:       final,
+		interrupted: interrupted,
+	})
+}
+
+type subagentRunOutcome struct {
+	final       string
+	interrupted *adk.InterruptInfo
+	err         error
+}
+
+func (e *Executor[M]) resolveRunOutcome(
+	ctx context.Context,
+	task *backgroundtask.Task,
+	control backgroundtask.ControlRequest,
+	cancelOutcome error,
+	run subagentRunOutcome,
+) (*backgroundtask.ExecutionResult, error) {
+	drainLostRace := control.Kind == backgroundtask.ControlDrain &&
+		errors.Is(cancelOutcome, adk.ErrExecutionEnded)
+	if !drainLostRace {
+		if result, controlErr, controlled := e.controlResult(ctx, task, control); controlled {
+			return result, controlErr
+		}
 	}
-	if interrupted != nil {
-		return e.interruptResult(ctx, task, interrupted)
+	if run.err != nil {
+		if errors.Is(run.err, adk.ErrSessionBusy) {
+			return &backgroundtask.ExecutionResult{
+				Directive:  backgroundtask.ExecutionDirectiveYield,
+				Checkpoint: append([]byte(nil), task.Checkpoint...),
+			}, nil
+		}
+		return nil, run.err
+	}
+	if run.interrupted != nil {
+		return e.interruptResult(ctx, task, run.interrupted)
 	}
 	return &backgroundtask.ExecutionResult{
 		Status: backgroundtask.StatusCompleted,
-		Data:   []byte(final),
+		Data:   []byte(run.final),
 	}, nil
 }
 
@@ -604,6 +633,28 @@ func decodeResumeTargets(data []byte) (map[string]any, error) {
 	return targets, nil
 }
 
+func requestCancelAndWait(
+	control backgroundtask.ControlRequest,
+	cancelRun adk.AgentCancelFunc,
+	drainCancelTimeout time.Duration,
+) error {
+	cancelOptions := []adk.AgentCancelOption{adk.WithRecursive()}
+	if control.Kind == backgroundtask.ControlDrain {
+		cancelOptions = append(cancelOptions,
+			adk.WithAgentCancelMode(adk.CancelAfterChatModel|adk.CancelAfterToolCalls))
+		if drainCancelTimeout > 0 {
+			cancelOptions = append(cancelOptions,
+				adk.WithAgentCancelTimeout(drainCancelTimeout))
+		}
+	} else {
+		cancelOptions = append(cancelOptions, adk.WithAgentCancelMode(adk.CancelImmediate))
+	}
+	// accepted is intentionally ignored: a rejected cancellation still returns
+	// a handle whose Wait reports the terminal run outcome.
+	handle, _ := cancelRun(cancelOptions...)
+	return handle.Wait()
+}
+
 // handleRunError translates agent event and output materialization errors into
 // durable task lifecycle outcomes when a control request is active.
 func (e *Executor[M]) handleRunError(
@@ -611,6 +662,7 @@ func (e *Executor[M]) handleRunError(
 	iter *adk.AsyncIterator[*adk.TypedAgentEvent[M]],
 	task *backgroundtask.Task,
 	controlRequests <-chan backgroundtask.ControlRequest,
+	cancelOutcomes <-chan error,
 	err error,
 ) (*backgroundtask.ExecutionResult, error) {
 	control := pollControl(controlRequests)
@@ -628,16 +680,8 @@ func (e *Executor[M]) handleRunError(
 			}
 		}
 	}
-	if result, controlErr, controlled := e.controlResult(ctx, task, control); controlled {
-		return result, controlErr
-	}
-	if errors.Is(err, adk.ErrSessionBusy) {
-		return &backgroundtask.ExecutionResult{
-			Directive:  backgroundtask.ExecutionDirectiveYield,
-			Checkpoint: append([]byte(nil), task.Checkpoint...),
-		}, nil
-	}
-	return nil, err
+	cancelOutcome := waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+	return e.resolveRunOutcome(ctx, task, control, cancelOutcome, subagentRunOutcome{err: err})
 }
 
 func waitForControl(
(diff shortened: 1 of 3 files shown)


adk/backgroundtask/subagent/subagent.go
```diff
@@ -504,8 +508,17 @@ func (e *Executor[M]) Execute(
...
+		cancelErr := waitForCancelOutcome(ctx, cancelOutcomes)
+		// A drain can lose the race with natural completion. In that case no
+		// resumable checkpoint is expected, so preserve the completed result.
+		if control.Kind != backgroundtask.ControlDrain ||
+			!errors.Is(cancelErr, adk.ErrExecutionEnded) {
+			if controlResult, controlErr, controlled := e.controlResult(ctx, task, control); controlled {
+				return controlResult, controlErr
+			}
+		}
 	}
```
shentongmartin, line 529: `ErrExecutionEnded` covers "normal completion **or a fatal error**" (cancel.go:204-205), but the comment here states the invariant generally. It holds for normal completion only.

A fatal event error goes through `handleRunError` (:492 -> :622), which polls the same channel at :629 but doesn't consult `cancelOutcomes` — so a drain losing the race to a *crash* still reproduces the original symptom:

```
execute err = backgroundtask: drain checkpoint unavailable: runner checkpoint is missing
final status="running"
```

To be clear: **this is pre-existing** — I get byte-identical output on `24279d81` and on this branch. Not asking you to grow the PR. But could you either narrow the comment to normal-completion-only and note the fatal case as known-open (follow-up issue), or extend the guard to `handleRunError`?

If you extend it: falling through to the return at :526-529 would yield `StatusCompleted` with empty `Data`, turning a crash into a success. The fatal case has to stay a failure.

Separately, the condition reads as a double negative for a positive concept. Naming the predicate makes the invariant self-documenting — this suggestion also carries the narrowed comment:

```suggestion
	control := pollControl(controlRequests)
	if control.Kind != "" {
		cancelOutcome := waitForCancelOutcome(ctx, cancelOutcomes)
		// A drain can lose the race with the run's own completion.
		// ErrExecutionEnded means the stream drained with no interrupt, so no
		// resumable checkpoint exists and the drain must defer to the run's
		// outcome instead of failing closed. Note this sentinel also covers
		// fatal errors, which reach this point only via handleRunError.
		drainLostRace := control.Kind == backgroundtask.ControlDrain &&
			errors.Is(cancelOutcome, adk.ErrExecutionEnded)
		if !drainLostRace {
			if controlResult, controlErr, controlled := e.controlResult(ctx, task, control); controlled {
				return controlResult, controlErr
			}
		}
	}
```

Also renamed `cancelErr` -> `cancelOutcome`: it's the cancel *operation's* verdict, not the agent's business error, and `nil` here means "cancel succeeded" — easy to misread as "no error occurred".

adk/backgroundtask/subagent/subagent.go
```diff
@@ -461,9 +462,12 @@ func (e *Executor[M]) Execute(
 			} else {
 				cancelOptions = append(cancelOptions, adk.WithAgentCancelMode(adk.CancelImmediate))
 			}
-			if handle, accepted := cancelRun(cancelOptions...); accepted {
-				_ = handle.Wait()
+			handle, _ := cancelRun(cancelOptions...)
```
shentongmartin: Worth a line here, because this is where the bug actually was: a *rejected* cancel still returns a non-nil handle whose `Wait` reports the correct terminal sentinel (cancel.go:872, 874, 885, 888, 897). The old `if accepted` skipped `Wait()` in precisely the racing case this PR fixes.

Without a note, someone may "restore" the check as a safety improvement and reintroduce the bug.

```go
// accepted is intentionally ignored: a cancel rejected because execution
// already finalized still returns a handle whose Wait reports the outcome.
```

adk/backgroundtask/subagent/subagent.go
```diff
@@ -461,9 +462,12 @@ func (e *Executor[M]) Execute(
...
+			if handle == nil {
+				cancelOutcomes <- nil
+				return
 			}
```
shentongmartin: This branch is unreachable today — every return path in `buildCancelFunc` constructs a handle. Fine to keep defensively, but worth saying so, otherwise readers go hunting for the nil-returning path.

adk/backgroundtask/subagent/subagent.go
```diff
@@ -654,6 +667,15 @@ func waitForControl(
...
 }
 
+func waitForCancelOutcome(ctx context.Context, outcomes <-chan error) error {
+	select {
+	case err := <-outcomes:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
shentongmartin: This blocking wait is safe only because it's reached *after* the iterator drains, which guarantees `markDone` fired -> `doneChan` closed -> `waitForCompletion` returns (cancel.go:851-864, :969-976). Load-bearing but invisible: a future refactor moving this call earlier reintroduces a deadlock. Worth stating on the function.

(Also note it deliberately omits `waitForControl`'s 100ms fallback — correct, since the outcome is guaranteed here rather than best-effort.)

adk/backgroundtask/subagent/subagent_test.go
```diff
@@ -1446,6 +1446,63 @@ func TestStopControlWinsOverLateFinalMessage_BitsUT(t *testing.T) {
...
+	}()
+	<-agent.started
+
+	closeDone := make(chan error, 1)
+	go func() {
+		closeCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+		closeDone <- manager.Close(closeCtx)
+	}()
+	time.Sleep(20 * time.Millisecond)
```
shentongmartin: This constructs the race probabilistically. On a loaded runner the drain can land outside the window and the test passes *without* exercising the fixed path — a silent false-pass, which is worse than a flake.

A handshake would make it deterministic: have the agent block until it observes the drain-induced cancellation, or expose a test-only hook that fires once `requestControlWithReason(ControlDrain, ...)` has queued.

I ran 40 jittered offsets (0-7ms) and behavior was stable, so this is test durability, not correctness.

adk/backgroundtask/subagent/subagent_test.go
```diff
@@ -1446,6 +1446,63 @@ func TestStopControlWinsOverLateFinalMessage_BitsUT(t *testing.T) {
...
+	close(agent.release)
+
+	require.NoError(t, <-executeDone)
```
shentongmartin: This is where the bug surfaced, but the assertion doesn't name what it guards, so an unrelated future error would read as the same regression:

```suggestion
	execErr := <-executeDone
	require.NotErrorIs(t, execErr, backgroundtask.ErrDrainCheckpointUnavailable,
		"drain that lost the race must not fail closed")
	require.NoError(t, execErr)
```

Separately: nice call asserting the exact `"late completion"` payload at :1497 and the checkpoint's *absence* at :1503 — that second one is easy to omit and it pins half the invariant.

shentongmartin (commented):
Root cause diagnosis is precise, and consuming `CancelHandle.Wait()` is the right mechanism. The subtle part — that the old `if accepted { handle.Wait() }` was itself what discarded the outcome in exactly the racing case — is easy to miss.

Verified locally: new test red on `24279d81` (3/3), green on `c744606a` (5/5); 40 jittered race iterations deterministic; `-race` clean; `./adk/backgroundtask/...` green; lint 0 issues. The fail-closed invariant for drains that *do* take effect holds.

Left line comments. One needs resolving before merge (the comment at :514 asserts an invariant that holds for only one of two paths); the rest are nits.

Two items I couldn't anchor inline because they're outside this diff:

- **subagent.go:265-267** — `SupportsDrain`'s doc still says drain "captures an ADK Runner checkpoint before returning a suspended result." That's now conditional; a drain that loses the race returns completed with no checkpoint.
- **subagent.go:622-654** — `handleRunError` polls the same `controlRequests` channel at :629 and calls `controlResult` at :644, but never consults `cancelOutcomes`. See the :511-522 comment for why this matters.

adk/backgroundtask/subagent/subagent_test.go
N3kox: require.NoError 已经覆盖了这里所说的需要补充的判断

adk/backgroundtask/subagent/subagent_test.go
```diff
@@ -1446,6 +1466,56 @@ func TestStopControlWinsOverLateFinalMessage_BitsUT(t *testing.T) {
...
+	executor := newTestExecutor(t, store)
+	task := &backgroundtask.Task{Spec: backgroundtask.Spec{ID: "task"}}
+	agent := &cancelThenMessageAgent{
+		name: "worker", started: make(chan struct{}), release: make(chan struct{}),
+	}
+	cancelOption, cancelRun := adk.WithCancel()
+	runner := adk.NewRunner(context.Background(), adk.RunnerConfig{Agent: agent})
+	iter := runner.Run(context.Background(), nil, cancelOption)
+	<-agent.started
+	close(agent.release)
```
shentongmartin, line 1486: This replaces the E2E test with a direct call on the extracted helper, and that loses coverage of the wiring where the original bug actually lived.

**Mutation check.** I severed only `Execute`'s wiring, leaving `resolveRunOutcome` and everything else untouched — a full revert of this PR's user-visible behavior:

```go
control := pollControl(controlRequests)
var cancelOutcome error // MUTANT: outcome never consulted
return e.resolveRunOutcome(ctx, task, control, cancelOutcome, subagentRunOutcome{...})
```

Result:

```
ok  github.com/cloudwego/eino/adk/backgroundtask/subagent  1.416s
```

Green — the bug ships and CI says nothing. The same mutant against an E2E shape:

```
outcome distribution over 40 iterations: map[running/execErr:40]   (correct code: map[completed:40])
execute err = backgroundtask: drain checkpoint unavailable: runner checkpoint is missing
final status="running"
```

To be clear, your new tests are not weak — I also mutated `drainLostRace` to a constant `false` and both this test and the `handleRunError` subtest correctly **FAIL**. They guard the *decision logic* properly. The gap is strictly the *wiring*, which is exactly where the bug was: `handle.Wait()`'s result being discarded on its way to a correct decision function.

The consequence worth weighing: the deleted test was red-on-base / green-on-head. This one passes on `24279d81` too, since `resolveRunOutcome` didn't exist there to break. So the PR no longer contains a test that would have caught the bug it fixes.

Suggest keeping these unit tests and restoring **one** E2E case alongside. I verified this drop-in FAILs on the severed-wiring mutant and PASSes 5/5 on your current head:

```go
func TestDrainControlDefersToNaturalCompletionE2E_BitsUT(t *testing.T) {
	store := adksession.NewInMemoryStore[*schema.Message](nil)
	agent := &cancelThenMessageAgent{
		name: "worker", started: make(chan struct{}), release: make(chan struct{}),
	}
	executor, err := NewExecutor(&ExecutorConfig[*schema.Message]{
		SessionStore: store, CheckPointStore: store, DrainCancelTimeout: time.Second,
	})
	require.NoError(t, err)
	require.NoError(t, executor.Register(
		agent.name, &AgentRegistration[*schema.Message]{Agent: agent},
	))
	registry := backgroundtask.NewExecutorRegistry()
	require.NoError(t, registry.Register(executor))
	manager := mustNewBackgroundManager(
		t, context.Background(), &backgroundtask.Config{Executors: registry},
	)
	task, err := Submit(context.Background(), manager, &SubmitRequest[*schema.Message]{
		SubAgentName: agent.name, Input: textInput("work"), Description: "work",
		SessionID: "parent",
	})
	require.NoError(t, err)

	executeDone := make(chan error, 1)
	go func() { executeDone <- manager.Execute(context.Background(), task.Spec.ID) }()
	<-agent.started

	closeDone := make(chan error, 1)
	go func() {
		closeCtx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
		defer cancel()
		closeDone <- manager.Close(closeCtx)
	}()
	time.Sleep(20 * time.Millisecond)
	close(agent.release)

	execErr := <-executeDone
	require.NotErrorIs(t, execErr, backgroundtask.ErrDrainCheckpointUnavailable,
		"drain that lost the race must not fail closed")
	require.NoError(t, execErr)
	require.NoError(t, <-closeDone)

	completed, err := manager.Get(context.Background(), task.Spec.ID)
	require.NoError(t, err)
	require.Equal(t, backgroundtask.StatusCompleted, completed.Status)
	require.Equal(t, "late completion", string(completed.ResultData))
}
```

This is where the `NotErrorIs` line earns its keep: on the mutant it's what names `ErrDrainCheckpointUnavailable` instead of reporting a generic non-nil error. Swap it for plain `require.NoError` if you prefer — the E2E coverage is the part that matters.

On the 20ms sleep: my earlier durability caveat still applies (on a loaded runner the drain can land outside the window). A handshake would be better, but a probabilistic E2E test beats none.

Also worth noting: putting the new fatal-error case inside the existing `TestHandleRunErrorControlOutcomes` table was the right call, and `require.Nil(t, result)` there is precisely the assertion that catches the "crash laundered into StatusCompleted" trap.

adk/backgroundtask/subagent/subagent.go
```diff
@@ -654,6 +698,25 @@ func waitForControl(
...
+// fully drained. Draining closes the cancellation lifecycle and guarantees that
+// a pending CancelHandle.Wait can complete.
+func waitForDrainCancelOutcome(
+	ctx context.Context,
+	control backgroundtask.ControlRequest,
+	outcomes <-chan error,
+) error {
+	if control.Kind != backgroundtask.ControlDrain {
+		return nil
+	}
```
shentongmartin: The early return only checks the control kind, so a **drain** control with a nil `outcomes` channel leaves one permanently-blocked case plus `ctx.Done()`. With a background context that blocks forever — confirmed by probe:

```
CONFIRMED: blocks forever on nil channel with drain control
```

Latent rather than live: the update threads a literal `nil` for `cancelOutcomes` at 6 test call sites (:1027, :1041, :1058, :1072, :1105, :1122), and those are safe *today* only because their control kind isn't `ControlDrain` — the two drain cases correctly pass a real channel. The safety of every `nil` therefore rests on an invariant no signature expresses.

Cheap to make explicit:

```suggestion
	if control.Kind != backgroundtask.ControlDrain || outcomes == nil {
		return nil
	}
```

Alternatively drop the `nil`s in tests in favour of a closed or buffered channel, so the production and test shapes agree.

shentongmartin (commented):
Re-reviewed at `1bfe2ea6` (base `24279d81` unchanged).

**The blocker is fixed, and fixed the more thorough way.** You took the invasive option — wiring the guard into `handleRunError` rather than just narrowing the comment — and the `resolveRunOutcome` extraction is better structured than what I proposed: unifying the completion, interrupt and error paths through one decision point removes the whole "two paths disagree" class rather than the single instance I found. Confirmed via the v1 attack suite:

```
TestAttack_DrainLosesRaceToFatalError
  before: status="running" + ErrDrainCheckpointUnavailable
  now:    status="failed"  resultError="model backend exploded"
```

The trap I flagged was avoided — the crash isn't laundered into `StatusCompleted`. All 5 nits addressed; dropping the dead `handle == nil` branch outright is better than the comment I suggested (verified: all 7 `AgentCancelFunc` return paths in cancel.go construct a non-nil handle). `./adk/backgroundtask/...` green across 7 packages, lint 0 issues, other two `SupportsDrain` implementers unaffected.

One new blocker introduced by the update — the deleted E2E test. Details on the test file; it's a test-shape regression, not a code defect, and ~40 lines closes it.

**Re your reply on the `NotErrorIs` nit:** you're right that `require.NoError` detects it — that was about diagnostic intent, not detection, and it's a fair judgment call. I accept it. Noting only that the assertion in question no longer exists on this branch, so the point resurfaces inside the E2E comment below; if you restore the test with plain `require.NoError`, that's fine by me.

Two smaller notes:

- **`ErrSessionBusy` relocation** — moving yield handling from `handleRunError` into `resolveRunOutcome` (subagent.go:534-541) means it's now evaluated on the `Execute` path too. Unreachable there in practice (`run.err == nil`), so behavior is preserved and the suite agrees — but worth a line in the PR description, since someone diffing `handleRunError` will wonder where the logic went.
- **Scope** — 2 files/+83 grew to 3 files/+174, all in service of the blocker plus the suggested refactor. Justified, but the description still describes the narrower original change.

adk/backgroundtask/subagent/subagent_test.go
N3kox: done，恢复了之前的测试

adk/backgroundtask/subagent/subagent.go
N3kox: 补了 nil 判断

shentongmartin (approved):

PR #1283, opened 2026-09-13
fix(adk): avoid checkpoint readback during subagent drain

shentongmartin:
## Summary
- mark Runner checkpoint persistence errors with stable, distinct sentinels
- require a successful same-attempt checkpoint write before suspending a durable background subagent
- fail cooperative drain persistence errors in place instead of rewinding to a historical task checkpoint
- trust a successful checkpoint write instead of downloading the checkpoint again before suspension

## Motivation
A drain previously performed a full checkpoint `Get` after `Set` only to verify existence. Large checkpoints can make this readback consume the shutdown deadline. The Runner reports failed checkpoint writes through its async event stream, and the subagent executor now records the successful `Set` acknowledgement for the exact task checkpoint key.

Historical `Task.Checkpoint` metadata is not used as a fallback for a failed cooperative drain: the current attempt may already have performed work or appended partial session events, so rewinding could duplicate work and make the event log inconsistent. Lease recovery remains available for unplanned worker loss.

## Tests
- `go test ./...`
- `go test -race ./...`
- `golangci-lint run --timeout 5m ./adk/backgroundtask/subagent/... ./adk/... ./internal/core/...`

### DIFF
--- a/adk/backgroundtask/subagent/subagent.go
+++ b/adk/backgroundtask/subagent/subagent.go
@@ -475,6 +475,18 @@ func (e *Executor[M]) Execute(
 		if event.Action != nil && event.Action.Interrupted != nil {
 			interrupted = event.Action.Interrupted
 		}
+		if errors.Is(event.Err, adk.ErrCheckpointSave) {
+			control := pollControl(controlRequests)
+			_ = waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+			if control.Kind == backgroundtask.ControlDrain {
+				return nil, fmt.Errorf(
+					"%w: %v",
+					backgroundtask.ErrDrainCheckpointUnavailable,
+					event.Err,
+				)
+			}
+			return nil, event.Err
+		}
 		if event.Err != nil && interrupted == nil {
 			return e.handleRunError(
 				ctx, iter, task, controlRequests, cancelOutcomes, event.Err,
@@ -673,14 +685,29 @@ func (e *Executor[M]) handleRunError(
 			errors.Is(err, adk.ErrStreamCanceled)) {
 		control = waitForControl(ctx, controlRequests)
 	}
+	var checkpointErr error
 	if control.Kind != "" {
 		for {
-			if _, open := iter.Next(); !open {
+			event, open := iter.Next()
+			if !open {
 				break
+			}
+			if errors.Is(event.Err, adk.ErrCheckpointSave) {
+				checkpointErr = errors.Join(checkpointErr, event.Err)
 			}
 		}
 	}
 	cancelOutcome := waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+	if checkpointErr != nil {
+		if control.Kind == backgroundtask.ControlDrain {
+			return nil, fmt.Errorf(
+				"%w: %v",
+				backgroundtask.ErrDrainCheckpointUnavailable,
+				checkpointErr,
+			)
+		}
+		return nil, checkpointErr
+	}
 	return e.resolveRunOutcome(ctx, task, control, cancelOutcome, subagentRunOutcome{err: err})
 }
 
@@ -749,7 +776,7 @@ func (e *Executor[M]) interruptResult(
 	}, nil
 }
 
-func (e *Executor[M]) controlResult(
+func (*Executor[M]) controlResult(
 	_ context.Context,
 	task *backgroundtask.Task,
 	control backgroundtask.ControlRequest,
@@ -765,14 +792,6 @@ func (e *Executor[M]) controlResult(
 			Error:  reason,
 		}, nil, true
 	case backgroundtask.ControlDrain:
-		if _, exists, err := e.checkPointStore.Get(
-			context.Background(), checkpointID(task.Spec.ID),
-		); err != nil || !exists {
-			if err == nil {
-				err = errors.New("runner checkpoint is missing")
-			}
-			return nil, fmt.Errorf("%w: %v", backgroundtask.ErrDrainCheckpointUnavailable, err), true
-		}
 		stateBytes, err := json.Marshal(checkpointState{
 			Sequence: nextCheckpointSequence(task.Checkpoint),
 		})
--- a/adk/backgroundtask/subagent/subagent.go
+++ b/adk/backgroundtask/subagent/subagent.go
@@ -744,17 +744,11 @@ func waitForDrainCancelOutcome(
 	}
 }
 
-func (e *Executor[M]) interruptResult(
-	ctx context.Context,
+func (*Executor[M]) interruptResult(
+	_ context.Context,
 	task *backgroundtask.Task,
 	interrupted *adk.InterruptInfo,
 ) (*backgroundtask.ExecutionResult, error) {
-	if _, exists, err := e.checkPointStore.Get(ctx, checkpointID(task.Spec.ID)); err != nil || !exists {
-		if err == nil {
-			err = errors.New("backgroundtask/subagent: runner checkpoint is missing")
-		}
-		return nil, err
-	}
 	state := checkpointState{
 		Sequence: nextCheckpointSequence(task.Checkpoint),
 	}
--- a/adk/backgroundtask/subagent/subagent.go
+++ b/adk/backgroundtask/subagent/subagent.go
@@ -692,8 +692,8 @@ func (e *Executor[M]) handleRunError(
 			if !open {
 				break
 			}
-			if errors.Is(event.Err, adk.ErrCheckpointSave) {
-				checkpointErr = errors.Join(checkpointErr, event.Err)
+			if checkpointErr == nil && errors.Is(event.Err, adk.ErrCheckpointSave) {
+				checkpointErr = event.Err
 			}
 		}
 	}
--- a/adk/runner.go
+++ b/adk/runner.go
@@ -85,6 +85,9 @@ type CheckPointStore = core.CheckPointStore
 
 type CheckPointDeleter = core.CheckPointDeleter
 
+// ErrCheckpointSave identifies a Runner checkpoint persistence failure.
+var ErrCheckpointSave = errors.New("adk: checkpoint save failed")
+
 type TypedRunnerConfig[M MessageType] struct {
 	Agent           TypedAgent[M]
 	EnableStreaming bool
@@ -964,7 +967,12 @@ func typedRunnerHandleIterImpl[M MessageType](enableStreaming bool, store CheckP
 			return
 		}
 		if err := saveRunnerCheckpoint(enableStreaming, store, ctx, *checkPointID, info, sig, sessionState); err != nil {
-			gen.Send(&TypedAgentEvent[M]{Err: fmt.Errorf("%s: %w", errLabel, err)})
+			gen.Send(&TypedAgentEvent[M]{Err: fmt.Errorf(
+				"%w: %s: %w",
+				ErrCheckpointSave,
+				errLabel,
+				err,
+			)})
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


adk/backgroundtask/subagent/subagent.go
```diff
@@ -765,14 +786,6 @@ func (e *Executor[M]) controlResult(
 			Error:  reason,
 		}, nil, true
 	case backgroundtask.ControlDrain:
```
N3kox, line 846: **[P0 / must answer] Drain without a checkpoint write yields a "ghost suspension" that cannot be resumed.**

Removing the pre-suspension readback here means that whenever the state "drain requested, but the runner checkpoint was never written" occurs, `controlResult` returns `StatusSuspended` with no error — the task is advertised as a resumable suspension while pointing at a checkpoint that does not exist.

The load-bearing question is in `runner.go`: the cancel path only calls `saveCheckpointNow` when `cancelErr.interruptSignal != nil && checkPointID != nil`. A drain uses safe-point cancel modes (`CancelAfterChatModel|CancelAfterToolCalls`). **Please prove that a drain-mode cancel that resolves to `StatusSuspended` always carries a non-nil `interruptSignal`.** If any path can leave it nil, no checkpoint is written *and* no `ErrCheckpointSave` event is emitted, yet this branch still returns `StatusSuspended` — exactly the silent-corruption case the removed readback used to catch.

Verified locally (this PR branch vs. base `60ad5c6f`): once that state occurs, `runner.Resume` fails hard with `checkpoint[...] not exist`, whereas the base fails closed with `ErrDrainCheckpointUnavailable`. So this is a behavioral regression.

Suggestion: provide the proof that a drain always carries a non-nil interrupt signal, add a drain-without-signal fail-closed test, or keep a cheap existence check (see the note on `checkpointSaveError`).

adk/backgroundtask/subagent/subagent.go
```diff
@@ -673,14 +685,29 @@ func (e *Executor[M]) handleRunError(
...
 	cancelOutcome := waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+	if checkpointErr != nil {
+		if control.Kind == backgroundtask.ControlDrain {
+			return nil, fmt.Errorf(
+				"%w: %v",
+				backgroundtask.ErrDrainCheckpointUnavailable,
+				checkpointErr,
+			)
+		}
+		return nil, checkpointErr
```
N3kox: **[P1] A user Stop is turned into `Failed` by a single checkpoint write failure.**

Per the new test "late checkpoint error overrides stop": when `control.Kind == ControlStop` and an `ErrCheckpointSave` is observed, this returns `nil, checkpointErr`, which the executor commits as `StatusFailed`.

The user already asked to discard this work and the checkpoint will never be read, yet a failed write to that unread checkpoint flips a semantically correct `StatusCanceled` into `StatusFailed` (which may trigger alerts / auto-retry / error reporting). Please justify why `StatusFailed` is more correct than honoring the requested cancel here; absent a strong reason, prefer returning `StatusCanceled` on the Stop path.

adk/backgroundtask/subagent/subagent.go
```diff
@@ -673,14 +685,29 @@ func (e *Executor[M]) handleRunError(
...
+				checkpointErr = event.Err
+			}
 		}
 	}
 	cancelOutcome := waitForDrainCancelOutcome(ctx, control, cancelOutcomes)
+	if checkpointErr != nil {
+		if control.Kind == backgroundtask.ControlDrain {
+			return nil, fmt.Errorf(
+				"%w: %v",
+				backgroundtask.ErrDrainCheckpointUnavailable,
```
N3kox: **[P1] Drain checkpoint failure is deferred to a different pod's resume.**

Returning `ErrDrainCheckpointUnavailable` here is treated as non-fatal by the manager (`executor.go` maps it through and `manager.go` does not fail Close on it), leaving the task in `StatusRunning` for recovery. But the checkpoint write did fail — the old readback failed fast in place, whereas this defers the failure to another pod's `beginRun -> runner.Resume`.

Please confirm that when the task is picked up again, resume restarts cleanly rather than faulting downstream on a missing/partial checkpoint (i.e. show this doesn't just move the crater downstream).

adk/runner.go
```diff
@@ -1310,10 +1330,14 @@ func (r *sessionTurnResult[M]) finalize(ctx context.Context) error {
 	// resume cannot load a checkpoint that points to a corrupt event log.
 	if r.pendingCheckpoint != nil && r.checkPointID != nil {
 		if r.persistErr != nil {
-			return fmt.Errorf("%s: skipped because session event persistence failed: %w", r.pendingCheckpoint.errLabel, r.persistErr)
+			return newCheckpointSaveError(
+				r.pendingCheckpoint.errLabel+
+					": skipped because session event persistence failed",
```
N3kox, line 1361: **[P1] `ErrCheckpointSave` conflates two failures with different semantics.**

Here a *session-event* persistence failure is wrapped under the *checkpoint-save* sentinel (see the new `TestExecutorInterruptSessionPersistenceFailureFailsWithoutCheckpoint`). A caller doing `errors.Is(err, ErrCheckpointSave)` can no longer distinguish:

- **checkpoint write failed** — usually a transient storage issue; a retry is likely to succeed;
- **session-event persistence failed** — the event log is corrupt; retrying is pointless or dangerous (the checkpoint would reference an incomplete event log).

Sharing one sentinel forces a one-size-fits-all retry policy downstream. Please confirm whether this merge is intentional; if the two must be distinguishable, consider a separate sentinel for the session-persistence failure.

adk/runner.go
```diff
@@ -85,6 +85,26 @@ type CheckPointStore = core.CheckPointStore
...
 
+// ErrCheckpointSave identifies a Runner checkpoint persistence failure.
+var ErrCheckpointSave = errors.New("adk: checkpoint save failed")
+
+type checkpointSaveError struct {
+	label string
+	cause error
+}
+
+func (e *checkpointSaveError) Error() string {
```
N3kox, line 103: **[P2] Cheaper existence check + document the store consistency assumption + error-text change.**

Three smaller but real points:

1. **Existence check**: the readback's cost is downloading a large blob. Does `CheckPointStore` have (or could it cheaply grow) an `Exists`/HEAD operation? If so, a cheap existence check preserves the catch-all guard without the payload cost — far smaller blast radius than rerouting correctness through the event stream.
2. **Store consistency assumption**: the new logic trusts durability once `Set` returns nil. Under an ack-before-durable or eventually-consistent store, this trusts a write that isn't durable. Please document the store semantics this PR assumes, somewhere integrators will see it. (Conversely, removing the readback also eliminates a false "missing" under eventual consistency — if that's an intended benefit, state and test it.)
3. **Error text**: `checkpointSaveError.Error()` changed the format from `"%s: %w"` to `"%s: %s: %v"`. Is any log scraping / alerting / test (including downstream integrators) matching the old string? Please grep consumers, not just the unit tests.

N3kox (commented):
Review focused on the correctness bet this PR makes: replacing the pre-suspension checkpoint *readback* (a catch-all existence guard) with *trust* in the Runner's `ErrCheckpointSave` event stream. That's only safe if every path that leaves the runner checkpoint absent surfaces as an `ErrCheckpointSave` event. Most of the change is reasonable, but a few paths don't obviously satisfy that invariant.

I verified the key concern locally by fetching this PR branch and running it against the base commit (`60ad5c6f`): when a drain resolves with no checkpoint written, `controlResult(Drain)` now returns `StatusSuspended` with no error, and a subsequent `runner.Resume` fails hard with `checkpoint[...] not exist`. The base code fails closed in the same scenario with `ErrDrainCheckpointUnavailable`. Details inline (Comment 1 is the load-bearing one).

adk/backgroundtask/subagent/subagent.go
shentongmartin: Fixed in `72c6d450`. I did not rely on `CancelError.interruptSignal`: timeout/streaming paths can legitimately have empty public interrupt contexts while still writing a resumable checkpoint. The subagent now wraps the Runner checkpoint store per attempt and records a successful `Set` acknowledgement for the exact task checkpoint key. `resolveRunOutcome` returns `Suspended` only when that acknowledgement was observed; otherwise it fails closed. `TestAttack_DrainControlWithoutCheckpointWriteFailsClosed` covers the no-write case, while the existing timeout/streaming resume tests cover successful writes without a readback.

adk/backgroundtask/subagent/subagent.go
shentongmartin: Fixed in `72c6d450`. A Stop now keeps `StatusCanceled` when the only late persistence error is `ErrCheckpointSave`; that checkpoint is intentionally unread after cancellation. Session-event persistence failures remain failures because they represent a distinct event-log integrity problem. The regression is covered by the renamed `late checkpoint error does not override stop` subtest.

adk/runner.go
shentongmartin: Fixed in `72c6d450`. Session-event failures now use the separate exported sentinel `ErrSessionEventPersistence`; they no longer match `ErrCheckpointSave`. Both wrappers preserve the original underlying cause through `Unwrap`. `TestSessionEventPersistenceErrorPreservesSentinelAndCause` and `TestRunnerSessionInterruptCheckpointSkippedOnPersistFailure` verify the distinction.

adk/backgroundtask/subagent/subagent.go
shentongmartin: Fixed in `72c6d450`. Your downstream-failure concern reproduced: an initial attempt with no durable task checkpoint was left Running, then failed on the next worker with `compatible checkpoint is required`. The subagent now returns the underlying persistence failure in that case, so the current manager commits `StatusFailed` immediately. `ErrDrainCheckpointUnavailable` is retained only when `Task.Checkpoint` already contains a prior durable recovery point. `TestExecutorDrainPropagatesCheckpointWriteFailureWithoutReadback` now asserts the fail-in-place behavior, and `late checkpoint error preserves prior drain checkpoint` covers the redispatch case.

adk/runner.go
shentongmartin: Addressed in `72c6d450`. Instead of adding an `Exists` API, the subagent uses a per-attempt wrapper to observe the successful `Set` acknowledgement for its exact checkpoint key, avoiding both a full `Get` and an interface-breaking requirement. The `CheckPointStore` contract now explicitly requires successful `Set` to mean durable and read-after-write visible. I also restored the prior error text (`<label>: <cause>`) while keeping `errors.Is(err, ErrCheckpointSave)` support; a repository-wide grep found no non-test consumer of the temporary sentinel-prefixed text.

adk/runner.go
shentongmartin: Additional policy update in `43b25fa5`: the sentinels remain distinct, and both persistence categories are terminal for a cooperative drain. In particular, `ErrSessionEventPersistence` no longer triggers fallback to a prior task checkpoint, because the failed attempt may already have durably appended a partial session timeline.

adk/backgroundtask/subagent/subagent.go
shentongmartin: Update after the final review (`43b25fa5`): fail-closed is now stricter. A cooperative drain without a successful checkpoint `Set` from that same attempt returns a normal execution error and the manager commits `StatusFailed`; it no longer returns `ErrDrainCheckpointUnavailable` or rewinds to an older `Task.Checkpoint`. `TestAttack_DrainControlWithoutCheckpointWriteFailsClosed` verifies both no suspension and no lease-retry sentinel.

adk/backgroundtask/subagent/subagent.go
shentongmartin: Correction to my previous reply: the historical-checkpoint redispatch policy was still unsafe and has been removed in `43b25fa5`. A prior `Task.Checkpoint` does not prove that its Runner payload is consistent with session events already appended by the current attempt. All cooperative drain persistence failures now fail the current task in place. Lease retry remains only for unplanned worker/lease loss, not as a known drain-failure fallback. `TestAttack_DrainCheckpointWriteFailureFailsCurrentAttempt` and the session-persistence control case cover this policy.

adk/runner.go
shentongmartin: Update after `43b25fa5`: the successful current-attempt `Set` acknowledgement is now the only condition that permits `StatusSuspended`. If that acknowledgement is absent, the cooperative drain fails the task; no historical `Task.Checkpoint` fallback is attempted. This keeps the no-readback optimization without mixing an old Runner snapshot with newer session events.

N3kox (approved):

PR #1287, opened 2026-09-14
feat(adk): support dynamic background task heartbeat interval

N3kox:
#### Check the PR title.

- [ ] This PR title match the format: \<type\>(optional scope): \<description\>
- [ ] The description of this PR title is user-oriented and clear enough for others to understand.
- [ ] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level. [User docs repo](https://github.com/cloudwego/cloudwego.github.io)

en:
zh(optional):


shentongmartin (approved):

shentongmartin:
Post-merge follow-up after a deeper adversarial pass: the happy path and existing CI look good, but I found two heartbeat lifecycle edge cases worth addressing in a follow-up PR.

1. **A blocking `HeartbeatInterval` callback can prevent a completed task from committing.**
   `HeartbeatInterval` is called synchronously in `Manager.heartbeat` (`adk/backgroundtask/executor.go:1005,1016`). After the executor returns, `Manager.Execute` closes `heartbeatStop` and waits for `heartbeatDone` before committing (`executor.go:953-970`). If the callback blocks on I/O/a lock/config lookup, the heartbeat goroutine cannot observe `stop` or `ctx.Done()`, so the completed execution hangs before its result is persisted.

   I reproduced this through the public `Manager.Execute` path with `TestAttack_ExecuteCompletionBlockedByIntervalProvider`.

2. **Any positive interval is accepted even when it exceeds the active-attempt lease.**
   `nextHeartbeatInterval` only falls back for `<= 0` (`adk/backgroundtask/manager.go:252-262`), and the first heartbeat is delayed by the full returned value. With `ActiveAttemptTimeout=20ms` and `HeartbeatInterval=1h`, a still-running task becomes `pending` before its first heartbeat. Another worker may then claim the same logical task, risking duplicate side effects and causing the original attempt's later commit to lose authority.

   I reproduced this with `TestAttack_IntervalLongerThanLeaseExpiresActiveExecution`.

Suggested follow-up:
- make the provider explicitly non-blocking/panic-free (ideally read an atomic duration updated out of band), and add a shutdown/commit regression test;
- define and enforce, where possible, a lease-safety invariant such as `interval + worst-case heartbeat latency < lease timeout`, or at least document it prominently and fail fast in configurations where both values are known;
- document that the timer now uses fixed-delay semantics (next interval starts after the previous heartbeat succeeds), unlike the previous ticker cadence.

The normal dynamic case is stable in stress testing, and nil/non-positive values correctly fall back to 10s; these concerns are specifically about lifecycle safety under blocking or lease-incompatible configuration.

PR #1294, opened 2026-09-16
feat(adk): expose IsLastAttempt to the ShouldRetry callback

### LINKED ISSUE #1288: [Feature] adk: expose "is last attempt" (or MaxRetries) in TypedRetryContext passed to ShouldRetry
## Problem

`TypedModelRetryConfig.ShouldRetry` receives a `TypedRetryContext` (`adk/retry_chatmodel.go:114-141`, v0.9.19) that carries `RetryAttempt`, but nothing that tells the callback whether this decision is the **last** one in the retry cycle.

Inside the loop the config's `MaxRetries` decides that, and it is applied *after* the callback returns:

- Generate path: `ShouldRetry` at `:433`; `if attempt >= r.config.MaxRetries { break }` at `:457-458`; `applyDecisionForRetry` only at `:460`.
- Stream path: `ShouldRetry` at `:611`; `if attempt < r.config.MaxRetries { applyDecisionForRetry(...) }` at `:644-645`.

So on the last attempt a decision with `Retry: true` is accepted, but `ModifiedInputMessages`, `PersistModifiedInputMessages`, `AdditionalOptions` and `Backoff` are silently discarded, and the loop falls through to `RetryExhaustedError` (`:473` / `:656`). The callback has no way to know that its decision is about to be dropped.

## Why it matters

A host SDK that layers policy on top of `ShouldRetry` (structured "will retry in N ms" telemetry, its own backoff, context compression via `ModifiedInputMessages`, side-effect-free termination) needs to branch on "is this the final slot" — e.g. not emit a retry-scheduled event, not sleep, not run an expensive compression whose output will never be used. Today it can only guess.

Note that returning `Retry: false` on the last attempt is not equivalent: it returns the raw error / message instead of `RetryExhaustedError`, so callers that classify exhaustion via `errors.Is(err, ErrExceedMaxRetries)` must still answer `Retry: true` on the final slot and then work around the discarded fields.

## Proposed API (additive)

```go
type TypedRetryContext[M MessageType] struct {
    RetryAttempt int
    // MaxRetries mirrors TypedModelRetryConfig.MaxRetries for this cycle.
    MaxRetries int
    // IsLastAttempt is true when a Retry:true decision will not produce another
    // model call; ModifiedInputMessages / AdditionalOptions / Backoff are ignored.
    IsLastAttempt bool
    // ... existing fields unchanged
}
```

Either field alone would be enough; `IsLastAttempt` is the least ambiguous and stays correct if the attempt numbering ever changes. `failover_chatmodel.go:45-56` already threads an equivalent `hasMoreAttempts` flag through context for the failover path, so this would make the retry path symmetric with it.

Population is a one-liner at each `retryCtx := &TypedRetryContext[M]{...}` site (`:426`, `:555`, `:604`): `IsLastAttempt: attempt >= r.config.MaxRetries`.

## Workaround today & its cost

The callback closes over the same `MaxRetries` value that was put into the config and tests `retryCtx.RetryAttempt > maxRetries`. This duplicates a config invariant in user code, depends on `RetryAttempt` being exactly `attempt+1`, and breaks silently if the retry config is wrapped/rewritten by another layer (e.g. composed with `FailoverChatModel`, or populated via `ModelRequest.ModelRetryConfig` in `handler.go:66-69`).

## Minimal reproduction (v0.9.19)

```go
cfg := &adk.ModelRetryConfig{
    MaxRetries: 1,
    ShouldRetry: func(ctx context.Context, rc *adk.RetryContext) *adk.RetryDecision {
        // attempt 2 is the last one; this Backoff and ModifiedInputMessages are dropped
        return &adk.RetryDecision{Retry: true, Backoff: 5 * time.Second,
            ModifiedInputMessages: compress(rc.InputMessages)}
    },
}
// Model always fails -> ShouldRetry is called with RetryAttempt=1 and RetryAttempt=2.
// Nothing in rc distinguishes the two; only the second decision is discarded.
```

vitolee-1018:
#### What type of PR is this?

feat

#### Check the PR title.

- [x] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level. [User docs repo](https://github.com/cloudwego/cloudwego.github.io)

feat(adk): 向 ShouldRetry 回调暴露 IsLastAttempt

en:

**Problem**

`TypedModelRetryConfig.ShouldRetry` has no way to know that the decision it is about to return is the last one of the retry cycle.

On the final attempt the loop hits `if attempt >= r.config.MaxRetries { break }` *before* `applyDecisionForRetry`, so a `Retry: true` decision is still accepted, but its `ModifiedInputMessages`, `PersistModifiedInputMessages`, `AdditionalOptions` and `Backoff` are silently discarded and the cycle ends with `RetryExhaustedError`.

Answering `Retry: false` instead is not equivalent: it returns the raw model error or output rather than `RetryExhaustedError`, so callers that classify exhaustion with `errors.Is(err, ErrExceedMaxRetries)` must still answer `Retry: true` on the final slot and then work around the dropped fields.

The only workaround today is to close over `MaxRetries` in user code and re-derive the invariant from `RetryAttempt`, which breaks silently if the retry config is wrapped or rewritten by another layer.

**Change**

Adds `IsLastAttempt bool` to `TypedRetryContext[M]`, populated with `attempt >= r.config.MaxRetries` at all three construction sites in `adk/retry_chatmodel.go`:

- `generateWithShouldRetry`, the generate loop
- `streamWithShouldRetry`, the branch where `Stream()` itself returns an error
- `streamWithShouldRetry`, after the stream output has been consumed

This also makes the retry path symmetric with the failover path, which already threads an equivalent `hasMoreAttempts` flag through `adk/failover_chatmodel.go`.

**Two deliberate choices, happy to change either**

1. `MaxRetries` is **not** mirrored into the context. The issue notes that either field alone would be enough, and mirroring `MaxRetries` would push the very config invariant the issue complains about back into user code.
2. The field is documented in terms of behaviour ("answering `Retry: true` will not produce another model call") rather than as "attempt equals MaxRetries", so the contract stays accurate if an early-exit path is ever added.

**Compatibility**

Purely additive. `TypedRetryContext` is only ever constructed by the framework and handed to the callback as a pointer, so no user code constructs it. It is not part of any checkpoint schema and is not serialized.

**Tests**

New `TestRetryChatModel_IsLastAttempt` in `adk/chatmodel_retry_test.go`, covering all three construction sites plus the boundary:

| subtest | covers |
| --- | --- |
| `Generate` | generate loop, `MaxRetries=2` |
| `StreamReturnsError` | `Stream()` returns an error |
| `StreamRejected` | stream succeeds, `ShouldRetry` rejects the output |
| `MaxRetriesZero` | `MaxRetries=0`, the single attempt must be reported as the last |

Each subtest asserts inside the callback that `IsLastAttempt` is true exactly on the final decision, and cross-checks that it was true exactly once.

Verified on this branch:

- `gofmt -l adk/` clean
- `go vet ./...` clean
- `go test -race -count=1 ./adk/...` pass
- Negative control: forcing `IsLastAttempt: false` at the three sites makes all four subtests fail, so the test is not vacuous.

**Base branch**

Targeted at `alpha/10`, since that is where the 0.10 line is being developed. `CONTRIBUTING.md` still points at `develop`, which no longer exists. Happy to retarget if that is wrong.

zh(optional):

问题：`ShouldRetry` 无法得知本次决策是不是重试周期里的最后一次。最后一轮里 `if attempt >= MaxRetries { break }` 发生在 `applyDecisionForRetry` 之前，所以 `Retry: true` 仍会被接受，但决策里的 `ModifiedInputMessages` / `PersistModifiedInputMessages` / `AdditionalOptions` / `Backoff` 会被静默丢弃。改成返回 `Retry: false` 并不等价：那样返回的是原始错误或输出，而不是 `RetryExhaustedError`，用 `errors.Is(err, ErrExceedMaxRetries)` 判断重试耗尽的调用方就识别不到了。

改法：在 `TypedRetryContext` 上新增 `IsLastAttempt`，并在三个构造点填充。纯新增字段，用户只读不构造，不进任何 checkpoint 序列化，兼容性零风险。没有镜像 `MaxRetries`，issue 也提到两个字段只需其一。

Fixes #1288

None attached. The new field carries its own GoDoc comment, and the user docs site documents `ModelRetryConfig` / `ShouldRetry` at the capability level without enumerating `RetryContext` fields, so there is nothing to update there. The agent skill reference in `cloudwego/eino-ext` does list the `RetryContext` fields; happy to send a follow-up PR there to keep it in sync if you would like.


vitolee-1018:
Friendly ping — no workflow run has been triggered on this PR yet, so there are no CI results for a reviewer to look at.

@shentongmartin @meguminnnnnnnnn could you approve the workflow run when you get a chance? The PR touches only two files under `adk/` (`retry_chatmodel.go` and its test) . And if you think the approach is wrong or would rather solve this a different way, happy to keep discussing it here.

Also recording the design question that came up offline, so it is not lost:

> Does the last round actually need `Backoff` / `ModifiedInputMessages`?

No — and this PR does not change that. The framework is right to drop them, and the retry semantics are untouched. The gap is on the callback side: it cannot tell whether those values are still worth computing. `Backoff` is cheap, but `ModifiedInputMessages: compress(...)` usually costs another model call, and on the final attempt that cost is paid for nothing. Telemetry has the same problem — a "will retry in N ms" event emitted on the final attempt is simply wrong, since no retry follows.

`adk/failover_chatmodel.go` already threads an equivalent `hasMoreAttempts` flag, so this only brings the retry path in line with the failover path.

Verified locally on this branch: `gofmt -l adk/` and `go vet ./...` clean, `go test -race -count=1 ./adk/...` passing, plus a negative control (forcing `IsLastAttempt: false` at the three sites makes all four new subtests fail).

PR #1297, opened 2026-09-16
fix(adk): stabilize in-memory filesystem query order

### LINKED ISSUE #1296: adk: InMemoryBackend query results are nondeterministic
## Problem

`InMemoryBackend` stores files in a Go map and returns query results in collection order:

- `LsInfo` iterates `b.files` and a `dirInfo` map;
- `GlobInfo` iterates `b.files`;
- multi-file `GrepRaw` appends worker results in channel arrival order.

As a result, identical backend contents and requests can produce different output ordering across calls or runs. This is especially undesirable for an agent-facing backend because it makes tool prompts, snapshots, caches, and the model's first-seen files nondeterministic.

## Expected behavior

For the same filesystem state and request, the in-memory backend should return results in a stable order:

- `LsInfo`: ascending by `Path`;
- `GlobInfo`: ascending by `Path`;
- `GrepRaw`: ascending by `Path`, then line number (with a deterministic final tie-breaker if needed).

## Proposed change

Sort results at the public method boundary, keeping the existing API and matching semantics unchanged. Add regression tests that insert files in different orders and verify stable output, including the parallel multi-file grep path and context results.

This would only touch `adk/filesystem/backend_inmemory.go` and its tests.

ammend:
## What this PR does

Makes `InMemoryBackend` query results deterministic:

- `LsInfo` sorts entries by path.
- `GlobInfo` sorts matches by path.
- `GrepRaw` sorts matches by path and then line number, including the parallel multi-file and context-line paths.

The implementation sorts at the public method boundary, so collection and matching behavior remain unchanged.

## Why

`InMemoryBackend` stores files in maps, and multi-file grep collects worker results by channel arrival order. Identical filesystem state and requests could therefore produce different tool output ordering across runs. Stable ordering improves reproducibility for agent prompts, snapshots, and caches.

Fixes #1296.

## Tests

Added a regression test that writes files in a deliberately unsorted order and verifies deterministic output for `LsInfo`, `GlobInfo`, and parallel `GrepRaw` with context lines.

- `go test ./adk/filesystem/...`
- `go test -race ./adk/filesystem/...`


hi-pender:
Reviewed the diff against the shell semantics the tools claim to emulate. The sort keys themselves look right: `ls` sorts by name by default, and POSIX pathname expansion sorts glob results by collation, so `LsInfo`/`GlobInfo` ascending by path matches convention. Sorting grep by `(path, line)` does deviate from `grep -r` (fts traversal order) and from ripgrep's default parallel output, but it is exactly what `rg --sort path` exists for, and the middleware applies `applyPagination(offset, headLimit)` on top of these results — offset paging over a nondeterministic order is broken regardless, so determinism is worth the deviation here. Worth putting that argument in the commit message; it is stronger than "reproducibility" alone.

Two things I think should be resolved before merge.

**(a) The new grep ordering is overridden one layer up**

`adk/middlewares/filesystem/filesystem.go:994` re-sorts the result of `GrepRaw` by **basename**:

```go
sort.SliceStable(matches, func(i, j int) bool {
    return filepath.Base(matches[i].Path) < filepath.Base(matches[j].Path)
})
```

So the order the model actually sees is basename first, and the new backend `(path, line)` order only acts as a tie-breaker: `/z/a.go` still comes before `/a/b.go`. The PR does achieve determinism (a stable sort over a now-deterministic input is deterministic), but not the path ordering the issue asks for, and not anything a shell would produce.

Suggest handling it in this PR: either change that comparator to full path + line, or drop it entirely now that the backend guarantees order. A test case with the same basename in different directories (`/a/x.txt` vs `/b/x.txt`) would pin this down — it is the case that currently slips through.

**(b) `GlobInfo` ordering now contradicts the tool description**

`adk/middlewares/filesystem/prompt.go:141` (and the corresponding line in `GlobToolDescChinese`) tells the model:

> Returns matching file paths sorted by modification time

That is no longer true after this change. Implementation and the description shipped to the model need to agree. I would keep the path sort and update both descriptions rather than switch to mtime: `ModifiedAt` is set from `time.Now()` on write, so files written in the same batch tie easily and mtime ordering would reintroduce the nondeterminism this PR is removing.

---

Minor notes, not blocking:

- `LsInfo`/`GlobInfo` use `sort.Slice` while grep uses `sort.SliceStable`. No ties are possible in the first two (paths come from map keys), so either works, but one style would be nicer.
- Sorting the single-file path fixes a real ordering wart beyond determinism: with `EnableMultiline`, `findMultilineMatches` can emit duplicated and non-increasing line numbers for overlapping matches, and the sort is what makes the output match grep's "ascending within a file" behavior. Worth calling out in the PR description.
- The `Backend` interface in `adk/filesystem/backend.go` documents no ordering contract, so only `InMemoryBackend` is ordered and custom backends will behave differently. Consider documenting the expected ordering on the interface methods.
- Unrelated and pre-existing, but adjacent to this code: `GrepRaw` holds `b.mu.RLock()` while calling `applyContext`, which takes `b.mu.RLock()` again per file (`backend_inmemory.go:499`). Recursive `RLock` on a `sync.RWMutex` deadlocks if a writer queues in between. Probably worth a separate issue.

ammend:
Addressed both blocking points in dcba785:

- The middleware now sorts grep matches by full path and then line, so pagination observes the same deterministic order even for custom backends. Added a regression case for `/z/a.txt` versus `/a/b.txt`.
- Updated both English and Chinese glob tool descriptions to state that results are sorted alphabetically by path.

I kept the middleware sort instead of removing it because `Backend` implementations are pluggable and may not provide ordered results.

The latest `eino-unit-test` failure was an unrelated 600-second timeout in the `adk` package; all filesystem tests in that run passed. The new push will trigger a fresh run.

hi-pender (approved):

